PR #16, opened 2022-09-08
[KOGITO-7899]-Integrate Kaniko into SWF Operator

desmax74:
See: https://issues.redhat.com/browse/KOGITO-7899

**Checklist**

- [ ] Have you verified tha tall the tests are passing?


api/v1alpha1/kogitoserverlessworkflow_types.go
```diff
@@ -308,20 +309,31 @@ type State struct {
...
-	Errors      []Error      `json:"errors,omitempty"`
-	KeepActive  bool         `json:"keepActive,omitempty"`
-	Auth        Auth         `json:"auth,omitempty"`
-	Events      *[]Event     `json:"events,omitempty"`
-	Functions   []Function   `json:"functions,omitempty"`
-	AutoRetries bool         `json:"autoRetries,omitempty"`
-	Retries     Retry        `json:"retries,omitempty"`
-	States      []State      `json:"states"`
+// KogitoSWFSpec defines the desired state of KogitoServerlessWorkflow
+type KogitoSWFSpec struct {
```
ricardozanini, line 313: Why changing the name here? `KogitoServerlessWorkflowSpec` seems way better.

api/v1alpha1/kogitoserverlessworkflow_types.go
```diff
@@ -331,18 +343,15 @@ type Endpoint struct {
...
-	ReadyStatusCondition    StatusCondition = "Ready"
-)
-
-// KogitoServerlessWorkflowStatus defines the observed state of KogitoServerlessWorkflow
-type KogitoServerlessWorkflowStatus struct {
-	Conditions StatusCondition    `json:"conditions,omitempty"`
-	Endpoints  []Endpoint         `json:"endpoints"`
-	Address    duckv1.Addressable `json:"address,omitempty"`
+// KogitoSWFStatus defines the observed state of KogitoServerlessWorkflow
+type KogitoSWFStatus struct {
```
ricardozanini, line 347: Same thing here. We can use an alias in the CRD to ksw, so we would do `oc get ksw` for short.

builder/action.go
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Imported from Camel-k
+ */
+package builder
```
ricardozanini, line 18: Please use my repo instead. Let it be the source of truth. This way, we can make it a standalone library to be used in the near future.

ricardozanini (changes requested):
I'd revert the naming, and the builder should be in the original repo. Let's turn that repo into the source of truth. In the future, we can figure out to which org we should push it.

api/v1alpha1/kogitoserverlessworkflow_types.go
desmax74: @ricardozanini I'd leaved the habit of long names with SpringFrameworkAndHisLongNamesForContextAbstractFactoryBeansAndWhatever  when I've encountered KieSpec in the KieCloudOperator, but I can revert this habit :)

builder/action.go
desmax74: Ok, my understand was to use the poc as example, I'll polish the pr using the repo as a library

api/v1alpha1/kogitoserverlessworkflow_types.go
```diff
@@ -308,20 +309,31 @@ type State struct {
...
+	Events       *[]Event     `json:"events,omitempty"`
+	Functions    []Function   `json:"functions,omitempty"`
+	AutoRetries  bool         `json:"autoRetries,omitempty"`
+	Retries      Retry        `json:"retries,omitempty"`
+	States       []State      `json:"states"`
+	CommonConfig CommonConfig `json:"commonConfig,omitempty"`
+}
+
+// CommonConfig variables used in the templates
+type CommonConfig struct {
```
davidesalerno, line 330: Let's add here a marker like a TODO to remove this from the Workflow spec since this is not part of the Workflow but are information needed for the deployment.

Otherwise if we would like to leave this details in this CRD it is probably better to use the Metadata for storing these details

api/v1alpha1/kogitoserverlessworkflow_types.go
```diff
@@ -331,41 +343,57 @@ type Endpoint struct {
...
-}
+type BuildStatusCondition string
+
+const (
+	BuildingStatusCondition BuildStatusCondition = "Building"
+	ReadyStatusCondition    BuildStatusCondition = "Ready"
+)
 
-// KogitoServerlessWorkflow is the Schema for the kogitoserverlessworkflows API
+// KogitoServerlessWorkflowApp is the Schema for the kogitoserverlessworkflows API
```
davidesalerno, line 376: Why is this renaming needed?

ricardozanini: I agree. `KogitoServerlessWorkflow` is the perfect and preferred naming.

constants/const.go
```diff
@@ -0,0 +1,29 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package constants
+
+const (
+	// ComponentName just a name to identify this package/component/application
+	ComponentName       = "kogito-builder"
+	KanikoVersion       = "1.9.0"
+	KanikoExecutorImage = "gcr.io/kaniko-project/executor:v" + KanikoVersion
```
ricardozanini, line 22: Are we using these constants here?

ricardozanini (commented):
Despite the naming, seems really good for a first iteration!

api/v1alpha1/kogitoserverlessworkflow_types.go
desmax74: ok

api/v1alpha1/kogitoserverlessworkflow_types.go
desmax74: ok

controllers/builder.go
```diff
@@ -0,0 +1,72 @@
...
+		build.Status.Phase != api.BuildPhaseError &&
+		build.Status.Phase != api.BuildPhaseFailed {
+		fmt.Printf("\nBuild status is %s", build.Status.Phase)
+		build, err = builder.FromBuild(build).WithClient(cli).Reconcile()
+		if err != nil {
+			fmt.Println("Failed to run test")
+			//panic(fmt.Errorf("build %v just failed", build))
+			return false, err
+		}
+		time.Sleep(10 * time.Second)
```
ricardozanini, line 68: Don't do this here. The example I left in the PoC was to mimic the controller reconciliation pattern. So you trigger a reconciliation here, each build state then will be triggered by the Kubernetes APIs, not sleep.

desmax74: @ricardozanini  Ok, but do we want a fire and forget behaviour instead to return the result of the build ?

api/v1alpha1/kogitoserverlessworkflow_types.go
```diff
@@ -357,15 +385,15 @@ type KogitoServerlessWorkflow struct {
 	Status KogitoServerlessWorkflowStatus `json:"status,omitempty"`
 }
 
-// KogitoServerlessWorkflowList contains a list of KogitoServerlessWorkflow
+// KogitoServerlessWorkflowAppList contains a list of KogitoServerlessWorkflow
 // +kubebuilder:object:root=true
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-type KogitoServerlessWorkflowList struct {
+type KogitoServerlessWorkflowAppList struct {
```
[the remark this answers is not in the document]
desmax74: Yep

controllers/builder.go
ricardozanini: Take a look at the Camel-K controller/builder package. You will see how they do.

Basically every time k8s does a reconciliation, this method will be called, and you advance the action automatically. You don't need this loop here.

constants/const.go
desmax74: Was used for the secret, but can be dropped for this first version

controllers/builder.go
desmax74: Ok

constants/const.go
```diff
@@ -0,0 +1,40 @@
...
+package constants
+
+const (
+	// ComponentName just a name to identify this package/component/application
+	ComponentName = "kogito-builder"
+	KanikoVersion = "1.9.0"
+
+	BUILD_SECRET_NAME       = "regcred"
+	DEFAULT_USERNAME        = "Kogito"
+	DEFAULT_DOCKER_REGISTRY = "https://quay.io/"
```
spolti, line 25: isn't better to use just quay.io?

spolti (approved):

constants/const.go
desmax74: yep

controllers/builder.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+	"github.com/davidesalerno/kogito-serverless-operator/constants"
+	"github.com/ricardozanini/kogito-builder/api"
+	"github.com/ricardozanini/kogito-builder/builder"
+	"github.com/ricardozanini/kogito-builder/client"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"time"
+)
+
+func Build(sourceSwf []byte, dockerFile []byte) (bool, error) {
```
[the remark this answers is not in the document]
desmax74: ok, but not for the demo

controllers/builder.go
```diff
@@ -0,0 +1,72 @@
...
+			Registry: api.RegistrySpec{
+				Insecure: true,
+			},
+			Timeout: &metav1.Duration{
+				Duration: 5 * time.Minute,
+			},
+		},
+	}
+
+	build, err := builder.NewBuild(platform, "greetings:latest", "kogito-test").
```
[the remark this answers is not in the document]
desmax74: Yep, here https://github.com/davidesalerno/kogito-serverless-operator/pull/16/files#diff-1a08e4164fa3e9ae3667ecc99281594e936d0ee69a9f30f71b3f652f22a058fdR87

README.md
```diff
@@ -12,6 +12,36 @@ operator-sdk-v1.22.2
...
+```
+configure docker registry with quay.io credentials or with your preferred docker registry
+
+Create a secret 
+```sh
+kubectl create secret docker-registry regcred --docker-server=<registry_url> --docker-username=<registry_username> --docker-password=<registry_password> --docker-email=<registry_email>
+```
+
+```sh
+kubectl create namespace kogito-builder
```
davidesalerno, line 42: Probably this line is duplicated

desmax74: yes, removed

PR #30, opened 2022-10-19
[KOGITO-8050]Verify that Kaniko is able to push on Kubernetes interna…

desmax74:
See: https://issues.redhat.com/browse/KOGITO-8050

…l registry

Signed-off-by: desmax74 <mdessi@redhat.com>

**Checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,37 @@ make manifests
 
 More information can be found via the [Kubebuilder Documentation](https://book.kubebuilder.io/introduction.html)
 
+
+#### Local development on Minikube with internal registry 
+Start the minikube with
+```sh
+minikube start --cpus 4 --memory 4096 --addons enable registry --insecure-registry "localhost:5000"
+```
+In the makefile change
+IMAGE_TAG_BASE to 
+```sh
+IMAGE_TAG_BASE ?= localhost:5000/kiegroup/kogito-serverless-operator
+```
+
+in the kustomization.yaml
+```sh
+DEFAULT_REGISTRY_REPO=<cluster_ip>/kiegroup
+```
+
+images.newName to
+
+```sh
+newName: localhost:5000/kiegroup/kogito-serverless-operator
+```
+
+##### TODO
+In the next iterations the localhost:5000 for the local development will be set at runtime with the ip of the cluster 
+
+DEFAULT_REGISTRY_REPO= in the kustomization.yaml or in the platform will be set using 
+
+https://github.com/ricardozanini/kogito-builder/blob/main/util/minikube/registry.go#L19
+
+
 ## License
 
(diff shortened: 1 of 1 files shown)


config/manager/kustomization.yaml
```diff
@@ -13,7 +13,7 @@ configMapGenerator:
   literals:
   - DEFAULT_BUILDER_RESOURCE_NAME=Dockerfile
   - DEFAULT_KANIKO_SECRET_DEFAULT=regcred
-  - DEFAULT_REGISTRY_REPO=quay.io/kiegroup
+  - DEFAULT_REGISTRY_REPO=registry.kube-system.svc.cluster.local/kiegroup
```
davidesalerno, line 16: This alias is working for pushing but seems not to work for pulling the image.
Try to use the cluster ip of the registry.

If using the registry cluster ip will work both for pushing and pulling the image, I propose simply to document it in a brand new DEV_ON_MINIKUBE md file into a _docs_ folder (to be created in the root of the project).

spolti: there might be a way to retrieve the registry address internally by querying the k8s api, it will differ from ocp, minikube and k8s.

desmax74: @spolti https://github.com/davidesalerno/kogito-serverless-operator/pull/30/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R181

spolti: not really, it only takes into account the minikube addon...

README.md
```diff
@@ -149,6 +149,37 @@ make manifests
...
+
+#### Local development on Minikube with internal registry 
+Start the minikube with
+```sh
+minikube start --cpus 4 --memory 4096 --addons enable registry --insecure-registry "localhost:5000"
+```
+In the makefile change
+IMAGE_TAG_BASE to 
+```sh
+IMAGE_TAG_BASE ?= localhost:5000/kiegroup/kogito-serverless-operator
```
davidesalerno, line 161: I don't understand why we need to push the operator into the minikube internal registry.

Don't we only need to know the configurations needed to push the images produced by the operator via kaniko on the internal registry?

spolti: operator will ned to be available on operatorhub.
however, we still need a doc about running operator locally, but, that can be done with operator-sdk.

davidesalerno: In general I think that is is better to use the default operator image and explain only to the user to change the DEFAULT_REGISTRY_REPO value into the config map and restart the controller.

desmax74:
Replaced by https://github.com/davidesalerno/kogito-serverless-operator/pull/31

PR #47, opened 2022-11-24
KOGITO-8290 Setup GHA checks for files

radtriste:
https://issues.redhat.com/browse/KOGITO-8290


radtriste:
result of new job failures when something is missing: https://github.com/kiegroup/kogito-serverless-operator/actions/runs/3541156536/jobs/5945068616

.github/workflows/checks.yml
```diff
@@ -0,0 +1,89 @@
...
+
+jobs:
+  unit-tests:
+    concurrency:
+      group: kogito-serverless-operator-unit-tests-${{ github.head_ref }}
+      cancel-in-progress: true
+    timeout-minutes: 60
+    name: Unit tests And Coverage
+    runs-on: ubuntu-latest
+    steps:
```
desmax74, line 25: @radtriste can we add this packages installed for podman in the Container-builder ?  https://github.com/kiegroup/container-builder/blob/main/.github/workflows/test-go.yml#L13..L17

desmax74 (approved):

.github/workflows/checks.yml
radtriste: Is that mandatory ?
I mean there is no issue here

davidesalerno (approved):

go.sum
```diff
@@ -362,6 +367,7 @@ github.com/prometheus/procfs v0.6.0/go.mod h1:cz+aTbrPOrUb4q7XlbU9ygM+/jj0fzG6c1
 github.com/prometheus/procfs v0.7.3/go.mod h1:cz+aTbrPOrUb4q7XlbU9ygM+/jj0fzG6c1xBZuNvfVA=
 github.com/prometheus/procfs v0.8.0 h1:ODq8ZFEaYeCaZOJlZZdJA2AbQR98dSHSM1KW/You5mo=
 github.com/prometheus/procfs v0.8.0/go.mod h1:z7EfXMXOkbkqb9IINtpCn86r/to3BnA0uaxHdg830/4=
+github.com/ricardozanini/kogito-builder v0.0.0-20221011121056-f1b666583d6c/go.mod h1:WvF/rVowqH6vRN+g0T1p2cKYXrQNY0cJwFmr7sewFno=
```
spolti, line 370: it seems to be wrong.

go.mod
```diff
@@ -40,6 +40,7 @@ require (
 	github.com/Azure/go-autorest/logger v0.2.1 // indirect
 	github.com/Azure/go-autorest/tracing v0.6.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
+	github.com/bmatcuk/doublestar/v4 v4.4.0 // indirect
```
spolti, line 43: do you know from where this dep is coming from?

hack/addheaders.sh
```diff
@@ -15,6 +15,7 @@
 
 if ! hash addlicense 2>/dev/null; then
   go get -u github.com/google/addlicense
```
spolti, line 17: we can probably remove go get and keep only the go install this way go.mod don't gets updated.

we can use: 
```bash
go install github.com/google/addlicense@latest
```

or use another specific version.

go.sum
radtriste: removed

go.mod
radtriste: removed

spolti (approved):

PR #49, opened 2022-11-25
KOGITO-8177 Setup e2e GHA

radtriste:
https://issues.redhat.com/browse/KOGITO-8177

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -39,52 +39,23 @@ jobs:
         go-version: ${{ env.GO_VERSION }}
         cache: true
 
-    - name: start minikube
+    - name: Setup Minikube cluster
       if: ${{ matrix.cluster-type == 'minikube' }}
       id: minikube
       uses: medyagh/setup-minikube@latest
+      with:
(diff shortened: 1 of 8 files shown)


radtriste:
Looks like no e2e tests are launched whereas there should be at least one.
any idea ? @desmax74 @davidesalerno

radtriste:
at least now it works with KinD.
Gonna setup Minikube as well

radtriste:
ok so ... Issue happening is due to https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted which requires the `seccompProfile` security context to be defined, which we don't have in the `manager` deployment.
This is in fact not mandatory because usually you configure Kubernetes/OCP to have a default `seccompProfile` which would be applied automatically...

This is what I have done:
- Force Kubernetes version on KinD and Minikube to the latest OCP extended support version (I took 4.10 as 4.12 is arriving)
- Still check for PodSecurityPolicy warnings but ignore the warnings on seccompProfile missing and apply kube patch on the deployment so the operator can run

spolti (approved):

PR #56, opened 2023-01-20
[KOGITO-8298] Retrieving Kaniko task only if it is present into the P…

davidesalerno:
…latform CR

See: [KOGITO-8298](https://issues.redhat.com/browse/KOGITO-8298)

Signed-off-by: Davide Salerno <dsalerno@redhat.com>

**Checklist**

<details>
<summary>
</summary>

</details>

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -88,9 +88,21 @@ func (b *Builder) ScheduleNewBuildWithContainerFile(workflowName string, imageTa
 }
 
 func (b *Builder) ScheduleNewKanikoBuildWithContainerFile(workflowName string, imageTag string, workflowDefinition []byte, build api.BuildSpec) (*api.Build, error) {
-	ib := b.getImageBuilderForKaniko(workflowName, imageTag, workflowDefinition, build.Tasks[0].Kaniko)
+	task := findKanikoTask(build.Tasks)
+	ib := b.getImageBuilderForKaniko(workflowName, imageTag, workflowDefinition, task)
 	ib.WithTimeout(5 * time.Minute)
 	return b.BuildImage(ib.Build())
+}
+
+func findKanikoTask(tasks []api.Task) *api.KanikoTask {
+	if tasks != nil && len(tasks) > 0 {
+		for _, task := range tasks {
+			if task.Kaniko != nil {
+				return task.Kaniko
+			}
+		}
+	}
+	return &api.KanikoTask{}
 }
 
 func (b *Builder) ReconcileBuild(build *api.Build, cli client.Client) (*api.Build, error) {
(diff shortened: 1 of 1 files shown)


builder/builder.go
```diff
@@ -88,11 +88,23 @@ func (b *Builder) ScheduleNewBuildWithContainerFile(workflowName string, imageTa
 }
 
 func (b *Builder) ScheduleNewKanikoBuildWithContainerFile(workflowName string, imageTag string, workflowDefinition []byte, build api.BuildSpec) (*api.Build, error) {
-	ib := b.getImageBuilderForKaniko(workflowName, imageTag, workflowDefinition, build.Tasks[0].Kaniko)
+	task := findKanikoTask(build.Tasks)
```
spolti, line 91: I think it is better to return a bool flag on findKanikoTask as well, this way you can avoid proceeding with the code if the task does not exists.

ricardozanini: That makes sense. We could even call the method within the conditional.

davidesalerno: @spolti I followed you suggestion and added an additional bool return value that I use to skip the code execution if no Kaniko task is found.

builder/builder.go
```diff
@@ -88,11 +89,26 @@ func (b *Builder) ScheduleNewBuildWithContainerFile(workflowName string, imageTa
 }
 
 func (b *Builder) ScheduleNewKanikoBuildWithContainerFile(workflowName string, imageTag string, workflowDefinition []byte, build api.BuildSpec) (*api.Build, error) {
-	ib := b.getImageBuilderForKaniko(workflowName, imageTag, workflowDefinition, build.Tasks[0].Kaniko)
+	task, found := findKanikoTask(build.Tasks)
```
spolti, line 92: this:
```
task, found := findKanikoTask(build.Tasks)
if !found {
	return nil, fmt.Errorf("Unable to find KanikoTask in Build CR for workflow %s", workflowName)
}
```

can be turned in something like this: 
```
if task, found := findKanikoTask(build.Tasks); !found {
	return nil, fmt.Errorf("Unable to find KanikoTask in Build CR for workflow %s", workflowName)
} else {
	ib := b.getImageBuilderForKaniko(workflowName, imageTag, workflowDefinition, task)
	ib.WithTimeout(5 * time.Minute)
	return b.BuildImage(ib.Build())
}
```

spolti: it does not change nothing, just sabe a few lines :)

davidesalerno: Saved one line of code ;-)

desmax74 (approved):

spolti (approved):

ricardozanini (approved):

spolti:
@radtriste do you want to take a look before we merge it?

radtriste (approved):

PR #59, opened 2023-01-26
KOGITO-8535 Deploy pipelines: Push built image

radtriste:
https://issues.redhat.com/browse/KOGITO-8535

On branches nightly, the version is set with `-snapshot` version so we just push the nightly image as it is built.
The default image name on branch will be nightly.

Once a release is done, the imageTagBase will be changed to prod one.
And we push all together.

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -19,7 +19,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/kiegroup/kogito-serverless-operator
+  newName: quay.io/kiegroup/kogito-serverless-operator-nightly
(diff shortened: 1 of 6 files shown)


config/manager/kustomization.yaml
```diff
@@ -19,7 +19,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/kiegroup/kogito-serverless-operator
+  newName: quay.io/kiegroup/kogito-serverless-operator-nightly
   newTag: 2.0.0-snapshot
```
spolti, line 23: shouldn't be latest?

radtriste: `latest` won't be accepted by the operator-sdk framework as a correct version ;)

spolti (approved):

PR #73, opened 2023-02-17
[KOGITO-8649] - Externalize application properties via ConfigMap

ricardozanini:

Now the `application.properties` file is externalized in the operator. 
See the demo: https://youtu.be/W-1MxG2CkH0

See https://issues.redhat.com/browse/KOGITO-8649

**Checklist**

<details>
<summary>
</summary>

</details>


controllers/profiles/object_creators.go
```diff
@@ -202,3 +208,54 @@ func ensureWorkflowSpecConfigMapMutator(workflow *operatorapi.KogitoServerlessWo
...
+	return func(object client.Object) controllerutil.MutateFn {
+		return func() error {
+			if kubeutil.IsObjectNew(object) {
+				return nil
+			}
+			original, err := workflowPropsConfigMapCreator(workflow)
+			if err != nil {
+				return err
+			}
+			cm := object.(*corev1.ConfigMap)
```
davidesalerno, line 243: We are 100% that the object is a ConfigMap but I'll double check the type assertion result in order to avoid a panic.

ricardozanini: It's better to panic rather than silently fail in this case (or just a log in the console) since a conversion error here is an implementation problem.

davidesalerno (approved):

controllers/profiles/reconciler_dev.go
```diff
@@ -254,28 +266,44 @@ func getDeploymentFailureReasonOrDefaultReason(deployment *appsv1.Deployment) st
...
 			deployment := object.(*appsv1.Deployment)
 			volumes := make([]v1.Volume, 0)
 			volumeMounts := make([]v1.VolumeMount, 0)
 
-			volumes = append(volumes, v1.Volume{
-				Name: configMapWorkflowDefVolumeNamePrefix,
-				VolumeSource: v1.VolumeSource{
-					ConfigMap: &v1.ConfigMapVolumeSource{
-						LocalObjectReference: v1.LocalObjectReference{Name: cm.Name},
+			volumes = append(volumes,
```
desmax74, line 277: In this function we are considering volumes empty, is better to save the volumes present, because if someone in the future call this function after the external resources we wipe all the volumes present in the deployment.

controllers/profiles/object_creators.go
```diff
@@ -202,3 +208,54 @@ func ensureWorkflowSpecConfigMapMutator(workflow *operatorapi.KogitoServerlessWo
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      getWorkflowPropertiesConfigMapName(workflow),
+			Namespace: workflow.Namespace,
+			Labels:    labels(workflow),
+		},
+		// we could use utils.NewJavaProperties, but this way is faster
+		Data: map[string]string{
+			quarkusConfigFileName: "# Immutable Workflow Properties \n" +
+				"quarkus.http.port=" + strconv.Itoa(defaultHTTPWorkflowPort) + "\n" +
+				"quarkus.http.host=0.0.0.0\n",
```
spolti, line 224: as the operator has access to the pod's ip, isn't better default to it?

controllers/profiles/reconciler_dev.go
```diff
@@ -76,16 +80,18 @@ func newDevProfileReconciler(client client.Client, logger *logr.Logger) ProfileR
 
 func newDevelopmentObjectEnsurers(support *stateSupport) *devProfileObjectEnsurers {
 	return &devProfileObjectEnsurers{
-		deployment:        newObjectEnsurer(support.client, support.logger, defaultDeploymentCreator),
-		service:           newObjectEnsurer(support.client, support.logger, defaultServiceCreator),
-		workflowConfigMap: newObjectEnsurer(support.client, support.logger, workflowSpecConfigMapCreator),
+		deployment:          newObjectEnsurer(support.client, support.logger, defaultDeploymentCreator),
+		service:             newObjectEnsurer(support.client, support.logger, defaultServiceCreator),
+		definitionConfigMap: newObjectEnsurer(support.client, support.logger, workflowDefConfigMapCreator),
```
spolti, line 85:
```suggestion
		workflowConfigMap: newObjectEnsurer(support.client, support.logger, workflowDefConfigMapCreator),
```

wdyt?

utils/java_properties.go
```diff
@@ -0,0 +1,116 @@
...
+	"strconv"
+	"strings"
+)
+
+const (
+	javaPropertyValidKeyRegExp = "^[a-zA-Z$_][a-zA-Z0-9$_\\.\\-]*$"
+)
+
+var _ JavaProperties = &javaProperties{}
+var validKeyRegExp = regexp.MustCompile(javaPropertyValidKeyRegExp)
```
spolti, line 28: Doesn't Golang have a properties api already in place?
https://pkg.go.dev/github.com/magiconair/properties
we could use it instead implement a new one,

ricardozanini: That's nice! Mine is better and lightweight. :D

But I guess it's a better option to use one that's better tested, I think. And since they don't use any other modules, I believe we can do it: https://github.com/magiconair/properties/blob/main/go.mod

controllers/profiles/object_creators.go
ricardozanini: no, since this IP can change, we might have more pods in the same workflow, etc. It's a lot of effort to a low value.

controllers/profiles/object_creators.go
```diff
@@ -202,3 +208,54 @@ func ensureWorkflowSpecConfigMapMutator(workflow *operatorapi.KogitoServerlessWo
...
+func workflowPropsConfigMapCreator(workflow *operatorapi.KogitoServerlessWorkflow) (client.Object, error) {
+	return &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      getWorkflowPropertiesConfigMapName(workflow),
+			Namespace: workflow.Namespace,
+			Labels:    labels(workflow),
+		},
+		// we could use utils.NewJavaProperties, but this way is faster
+		Data: map[string]string{
+			quarkusConfigFileName: "# Immutable Workflow Properties \n" +
```
desmax74, line 222: Why not load from a yaml file like other things created during the startup ? https://github.com/kiegroup/kogito-serverless-operator/tree/main/bundle/manifests

controllers/profiles/reconciler_dev.go
spolti: ok

controllers/profiles/reconciler_dev.go
```diff
@@ -34,13 +34,17 @@ import (
...
-	configMapWorkflowDefVolumeNamePrefix    = "wd"
+	configMapWorkflowDefVolumeName          = "workflow-definition"
 	configMapWorkflowDefMountPath           = "/home/kogito/serverless-workflow-project/src/main/resources/workflows"
-	requeueAfterFailure                     = 3 * time.Minute
-	requeueAfterFollowDeployment            = 10 * time.Second
-	requeueAfterIsRunning                   = 3 * time.Minute
+	// quarkusDevConfigMountPath mount path for application properties file in the Workflow Quarkus Application
+	//
+	// See: https://quarkus.io/guides/config-reference#application-properties-file
+	quarkusDevConfigMountPath    = "/home/kogito/serverless-workflow-project/src/main/resources"
```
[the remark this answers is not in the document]
ricardozanini: Yes, we need previsibility for what we are doing. And the `artifactId` won't do anything for users.

controllers/profiles/object_creators.go
ricardozanini: @desmax74, Yeah, I thought about this over the weekend. The manifests are added by the bundle to the image, correct? So we have a path where we can look for?

ricardozanini: This configMap in this PR, can't be deployed by the manager since it will be one unique per workflow. 

What I'm looking for is to have an `application.properties` file in the bundle to load here, so we avoid hardcoding the default properties.

I'll take a look.

desmax74: @ricardozanini a trick could be load from the manager an application properties with the keys and placeholders as a values, every workflow define the values of the keys and after the replacement add it in the configmap or do you think is too complicated ?

ricardozanini: Loading a file every time is not the way to go, IMO. I've kept a var with the default properties. It's faster and more straightforward than either pinging the API or loading from a file.

ricardozanini:
@radtriste @spolti @desmax74 can you please re-review. I've wrapped up.

desmax74 (approved):

radtriste (approved):
tested ok

PR #77, opened 2023-02-24
[KOGITO-8645] Review Conditions Status API

ricardozanini:
In this PR, we change the way we deal with conditions following the Kubernetes convention.

See: https://issues.redhat.com/browse/KOGITO-8645

**Checklist**

<details>
<summary>
</summary>

</details>


ricardozanini:
@radtriste @davidesalerno @spolti feel free to review.

config/manager/manager.yaml
```diff
@@ -34,6 +34,8 @@ spec:
         image: controller:latest
         name: manager
         securityContext:
+          seccompProfile:
```
radtriste, line 37: that one will create issues with older kubernetes ...

controllers/kogitoserverlessworkflow_controller.go
```diff
@@ -29,9 +29,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/kiegroup/kogito-serverless-operator/api"
```
radtriste, line 32: operatorapi ?

radtriste (commented):
Why not reusing directly https://github.com/kubernetes/apimachinery/blob/release-1.23/pkg/apis/meta/v1/types.go#L1432-L1492 ?

ricardozanini:
> Why not reusing directly https://github.com/kubernetes/apimachinery/blob/release-1.23/pkg/apis/meta/v1/types.go#L1432-L1492 ?

 Because I want the liberty to change it and not be impacted if they change their API.

config/manager/manager.yaml
ricardozanini: Yes, we need to update/create the docs with a minimum version in mind. This was causing problems for the e2e. We can change the tests, but I think we can afford to have this strict security set. wdyt?

api/condition_types.go
```diff
@@ -0,0 +1,102 @@
...
+// +kubebuilder:object:generate=true
+type Condition struct {
+	// Type condition for the given object
+	// +required
+	Type ConditionType `json:"type"`
+	// Status of the condition, one of True, False, Unknown.
+	// +required
+	Status v1.ConditionStatus `json:"status"`
+	// The last time this condition was updated.
+	LastUpdateTime metav1.Time `json:"lastUpdateTime,omitempty"`
```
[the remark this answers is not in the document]
ricardozanini: It's not a transition, actually. It's an update. 😅

config/manager/manager.yaml
radtriste: we would still break with OCP EUS :S

api/condition_types.go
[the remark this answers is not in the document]
radtriste: I was referring to the meta/v1 api :)

api/status_types.go
```diff
@@ -0,0 +1,345 @@
...
+// +kubebuilder:object:generate=true
+type Status struct {
+	// The latest available observations of a resource's current state.
+	// +optional
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	Conditions Conditions `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+	// The generation observed by the deployment controller.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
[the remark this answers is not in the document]
radtriste: see the meta/v1 spec. That `ObservedGeneration` is for each condition AFAIU

controllers/kogitoserverlessworkflow_controller.go
ricardozanini: The operatorapi is this one:

```
operatorapi "github.com/kiegroup/kogito-serverless-operator/api/v1alpha08"
```

It's the versioned one where we can change only the import and keep the alias.

In this case, `api` is the better name.

api/status_types.go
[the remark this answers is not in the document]
ricardozanini: Oh sure, depends on the implementation. On Knative they use it in the whole status since every reconciliation cycle they reevaluate every condition. We are following a similar approach. So it makes sense keeping in a higher level.

Also in the Deployment Status, same thing:

```yaml
status:
  observedGeneration: 3
  replicas: 1
  updatedReplicas: 1
  readyReplicas: 1
  availableReplicas: 1
  conditions:
    - type: Available
      status: 'True'
      lastUpdateTime: '2023-02-28T11:25:36Z'
      lastTransitionTime: '2023-02-28T11:25:36Z'
      reason: MinimumReplicasAvailable
      message: Deployment has minimum availability.
    - type: Progressing
      status: 'True'
      lastUpdateTime: '2023-02-28T11:55:42Z'
      lastTransitionTime: '2023-02-28T11:25:32Z'
      reason: NewReplicaSetAvailable
      message: ReplicaSet "greeting-57f9548ff4" has successfully progressed.
```

api/condition_types.go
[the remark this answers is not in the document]
ricardozanini: Oh ok! That's because we don't transition to the conditions, they are evaluated together every time. We could do it, but I don't see a benefit of doing it at the moment.

For instance, we have a `Built` condition that will be either `Unknow` or `True` in dev. We won't transition to it, since it's a static condition in the dev mode. The `LastUpdateTime` just clarifies that in both cases.

config/crd/bases/sw.kogito.kie.org_kogitoserverlessworkflows.yaml
```diff
@@ -28,7 +28,7 @@ spec:
     - jsonPath: .status.address.url
       name: Address
       type: string
-    - jsonPath: .status.condition
+    - jsonPath: .status.conditions[?(@.type=='Running')].status
```
radtriste, line 31: Could we instead give the latest condition type ?
Or call it `RUNNING` instead of Condition ?

Also, I thin the `REASON` needs to be changed
```
$ kget KogitoServerlessWorkflow
NAME       PROFILE   VERSION   ADDRESS   CONDITION   REASON
greeting             0.0.1               False       
```

radtriste:
When testing, I tried with the `config/samples/sw.kogito_v1alpha08_kogitoserverlessworkflow.yaml` sample.
It gets built but the built image is launched as a pod.

And I think that is also the problem reported by the e2e tests.

Seems to be a reconciler issue or ?

config/crd/bases/sw.kogito.kie.org_kogitoserverlessworkflows.yaml
ricardozanini: > Could we instead give the latest condition type ?

What do you mean?

Actually, I need to change this. It should be:

```
$ kget KogitoServerlessWorkflow
NAME       PROFILE   VERSION   ADDRESS   READY   REASON
greeting             0.0.1               False       
```

Condition > Ready. And the reason should be changed, as you pointed! Nice catch.

ricardozanini:
>It gets built but the built image is launched as a pod.

wdym? I'll test it locally. This entire reconciliation must be reviewed, actually. I'll open a JIRA.

> And I think that is also the problem reported by the e2e tests.

hmm it might be related to the timing. I'll take a look at the new snapshot image. Might be a problem there.

ricardozanini:
> And I think that is also the problem reported by the e2e tests.

It was failing because we removed the `seccompProfile`. I've removed the test and the namespace with the labels enforcing it.

ricardozanini:
> It gets built but the built image is launched as a pod.

I'll review this use case tomorrow.

radtriste:
gonna have a new run this afternoon

ricardozanini:
The prod profile is still failing. It builds but does nothing. I'll fix it tomorrow.

radtriste:
@

> The prod profile is still failing. It builds but does nothing. I'll fix it tomorrow.

It could be related to the issue on runtimes ...

ricardozanini:
@radtriste all done.

builder/config.go
```diff
@@ -46,23 +46,22 @@ func NewCustomConfig(platform operatorapi.KogitoServerlessPlatform) (map[string]
 		return nil, fmt.Errorf("unable to retrieve the namespace from platform %s", platform.Name)
 	}
 	customConfig[configKeyBuildNamespace] = platform.Namespace
-	if platform.Spec.BuildPlatform.Registry.Secret == "" {
```
desmax74, line 49: The registry secret was required by Kaniko

builder/config.go
```diff
@@ -46,23 +46,22 @@ func NewCustomConfig(platform operatorapi.KogitoServerlessPlatform) (map[string]
...
-		return nil, fmt.Errorf("unable to retrieve the registry credentials from platform %s", platform.Name)
-	}
+	// Registry Secret and Address are not required, the inner builder will use minikube inner registry if available
+	// TODO: we should review
 	customConfig[configKeyRegistrySecret] = platform.Spec.BuildPlatform.Registry.Secret
-	if platform.Spec.BuildPlatform.Registry.Address == "" {
```
desmax74, line 53: The registry address was required by kaniko

builder/config.go
desmax74, line 49: The registry secret was required by Kaniko

desmax74 (approved):

ricardozanini:
@desmax74 thanks for your review, I remembered to open a new JIRA for prod. Please see: https://issues.redhat.com/browse/KOGITO-8806

We won't require a registry anymore. Minikube has the capability to use its internal registry. For OCP, the default setup with BuildConfig also is capable to use their internal registry. We shouldn't require these parameters. See the e2e working perfectly now with prod.

utils/kubernetes/deployment.go
```diff
@@ -52,12 +52,12 @@ func IsDeploymentProgressing(deployment *appsv1.Deployment) bool {
...
 
-// GetDeploymentUnavailabilityReason returns a string explaining why the given deployment is unavailable. If empty, there's no replica failure.
+// GetDeploymentUnavailabilityMessage returns a string explaining why the given deployment is unavailable. If empty, there's no replica failure.
 // Note that the Deployment might be available, but a second replica failed to scale. Always check IsDeploymentAvailable.
-func GetDeploymentUnavailabilityReason(deployment *appsv1.Deployment) string {
+func GetDeploymentUnavailabilityMessage(deployment *appsv1.Deployment) string {
 	for _, condition := range deployment.Status.Conditions {
-		if condition.Type == appsv1.DeploymentReplicaFailure &&
-			condition.Status == v1.ConditionTrue {
+		if condition.Type == appsv1.DeploymentAvailable &&
```
davidesalerno, line 59: Shouldn't be the deployment in status ```DeploymentUnavailableReason```

radtriste (approved):

utils/kubernetes/deployment.go
ricardozanini: It might! I'll take a look at the docs again.

ricardozanini: Actually, no. But I improved this function to look for two conditions in the underlying deployment. This way we can provide a more precise error message/reason to a failure.

PR #78, opened 2023-02-27
[KOGITO-8643] Expose Workflow Dev Profile application endpoint extern…

davidesalerno:
…ally

See: [KOGITO-8643](https://issues.redhat.com/browse/KOGITO-8643)

If the workflow is using the Dev profile the Operator will change the Service Type to NodePort and it will add to the KogitoServerlessWorkflow CRD status an Endpoint with the NodePort used for this service

Let the service to be easily available outside Kubernetes when using the Dev profile

**Checklist**

<details>
<summary>
</summary>

</details>


controllers/profiles/ensurer.go
```diff
@@ -33,11 +33,22 @@ func newObjectEnsurer(client client.Client, logger *logr.Logger, creator objectC
...
+
 // objectEnsurer provides the engine for a ReconciliationState that needs to create or update a given Kubernetes object during the reconciliation cycle.
 type objectEnsurer struct {
-	client  client.Client
-	logger  *logr.Logger
-	creator objectCreator
+	client     client.Client
+	logger     *logr.Logger
+	creator    objectCreator
+	conditions []objectCreationCondition
```
ricardozanini, line 51: Instead, I think we should have ensurers per cluster. This way, we create the right `ensurer` once we start the reconciliation and avoid looping in these conditions. In a nutshell, we should use the builder to have the right structure based on the conditionals we want, and not during the reconciliation cycle.

controllers/profiles/reconciler_dev.go
```diff
@@ -81,19 +83,31 @@ func newDevProfileReconciler(client client.Client, logger *logr.Logger) ProfileR
 func newDevelopmentObjectEnsurers(support *stateSupport) *devProfileObjectEnsurers {
 	return &devProfileObjectEnsurers{
 		deployment:          newObjectEnsurer(support.client, support.logger, defaultDeploymentCreator),
-		service:             newObjectEnsurer(support.client, support.logger, defaultServiceCreator),
+		service:             newObjectEnsurer(support.client, support.logger, devServiceCreator),
+		route:               newConditionedObjectEnsurer(support.client, support.logger, defaultRouteCreator, onlyForOpenShiftClusters),
```
ricardozanini, line 87: Instead, we can have `newDevelopmentObjectEnsurersForOpenShift`.

The ensurer then can be a `network` and we inject either the nodePort ensurer or the route one.

ricardozanini (commented):
Usually, I don't review draft PR. But here, I believe that we should review these new "conditions" attributes and leave to the builder the logic to create them based on the platform conditional.

api/v1alpha08/kogitoserverlessworkflow_types.go
```diff
@@ -38,7 +38,12 @@ type KogitoServerlessWorkflowSpec struct {
 	States      []State      `json:"states"`
 }
 
-var _ api.ConditionsReader = &KogitoServerlessWorkflowStatus{}
+type Endpoint struct {
+	IP       string `json:"ip,omitempty"`
+	Port     int    `json:"port,omitempty"`
+	PortName string `json:"portName,omitempty"`
+	Protocol string `json:"protocol,omitempty"` // "TCP" or "UDP"; never empty
```
ricardozanini, line 45: We should consider openapi level validation here via kubebuilder annotations.

controllers/profiles/reconciler_dev.go
```diff
@@ -129,12 +160,20 @@ func (e *ensureRunningDevWorkflowReconciliationState) Do(ctx context.Context, wo
...
 	objs = append(objs, deployment)
 
-	service, _, err := e.ensurers.service.ensure(ctx, workflow, defaultServiceMutateVisitor(workflow))
+	service, _, err := e.ensurers.service.ensure(ctx, workflow, defaultServiceMutateVisitor(workflow), devProfileServiceMutateVisitor(workflow))
 	if err != nil {
 		return ctrl.Result{RequeueAfter: requeueAfterFailure}, objs, err
 	}
 	objs = append(objs, service)
 
+	if e.ensurers.network != nil {
```
ricardozanini, line 169: Instead, apply the "null pattern" here. In the creator, we can have an ensurer that does nothing and just return. We avoid adding conditionals to the code.

controllers/profiles/reconciler_dev.go
```diff
@@ -180,8 +220,10 @@ func (f *followDeployDevWorkflowReconciliationState) Do(ctx context.Context, wor
 	if kubeutil.IsDeploymentAvailable(deployment) {
 		workflow.Status.Manager().MarkTrue(api.RunningConditionType)
 		f.logger.Info("Workflow is in Running Condition")
+		f.enrichers.workflowStatus.Enrich(ctx, workflow)
```
ricardozanini, line 223: Is this a post-processing, on-exit state of some sort? If so, we can treat it as this or just do what we need within the state, hard to identify. I think it's a better approach so we don't struggle to debug a problem caused by something happening after we exit the state. If we do an on-exit or post-processing, we should make it clear in the state API "Do, CanReconcile, OnExit".. etc.

operator.yaml
```diff
@@ -688,6 +691,9 @@ spec:
               cluster:
                 description: Cluster what kind of cluster you're running (ie, plain
                   Kubernetes or OpenShift)
+                enum:
+                - Kubernetes
+                - OpenShift
```
ricardozanini, line 696: Is this in the status or spec? We should not expose it for users to change this value.

davidesalerno: It is in the latform CRD spec. Why don't we want to expose it?

Let's suppose we are on OpenShift but on a specific namespace the user wouldn't like to use some advanced features related to OpenShift because we are developing for a different target cluster, in this way we will let the users to be free to do that.

ricardozanini: IDK, this use case seems really weird to me

davidesalerno: We can add a validation check that on a non OpenShift cluster that behaviour won't be permitted. 
Anyway I'll track this in a separate task because it is not related to this activity.

operator.yaml
```diff
@@ -875,6 +881,9 @@ spec:
     - jsonPath: .status.address.url
       name: Address
       type: string
+    - jsonPath: .status.endpoints[?(@.portName == "NodePort")].port
+      name: Port
+      type: string
```
ricardozanini, line 886: I think `endpoint` should not be an array. Why it is? `Endpoint` can be the exposed URI and `Address` to respect Knative interface? Address can be internal, Endpoint can be external.

bundle/manifests/kogito-serverless-operator.clusterserviceversion.yaml
```diff
@@ -11,7 +11,7 @@ metadata:
             "name": "kogito-workflow-platform"
           },
           "spec": {
-            "cluster": "kubernetes",
+            "cluster": "Kubernetes",
```
spolti, line 14: maybe it is better to keep everything lower case, no?

operator.yaml
davidesalerno: I used an array since in the initial design notes you introduced the `endpoints` as a plural entity and I  was thinking that it is due to the fact that with multiple pods there will be multiple endpoints if the kube scheduler will spread the pods in different nodes (in a production scenario it could happen quite often).
Anyway we can refine the design accordingly to your concern if we think that the reported use case could be managed using a different approach.

davidesalerno: We problably get inspirated by Kubernetes for the `Endpoints` as a plural entity https://kubernetes.io/docs/reference/kubernetes-api/service-resources/endpoints-v1/ but if this doesn't fit our needs we can change it

README.md
```diff
@@ -26,7 +26,7 @@ You’ll need a Kubernetes cluster to run against. You can use:
 
 - [KIND](https://sigs.k8s.io/kind)
 - [MINIKUBE](https://minikube.sigs.k8s.io/)
-- [CRC](https://console.redhat.com/openshift/create/local)
+- [Openshift Local](https://console.redhat.com/openshift/create/local)
```
spolti, line 29:
```suggestion
- [Openshift](https://console.redhat.com/openshift/create/local) - Using CRC
```

README.md
```diff
@@ -53,7 +53,7 @@ minikube start [...] --driver podman
 
 ```sh
 make install
-```
+``
```
spolti, line 56:
```suggestion
```
```

Comment (spolti) [inline README.md:56]: missing one crasis.

File: README.md [hunk 11]
```diff
@@ -154,6 +161,7 @@ kubectl delete namespace kogito-workflows
 make undeploy
 ```
 
+
```
spolti, line 164: unneeded extra line?

api/v1alpha08/kogitoserverlessworkflow_types.go
```diff
@@ -17,6 +17,7 @@ package v1alpha08
 import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/pkg/apis"
```
spolti, line 20: maybe re-organize the imports

config/rbac/kustomization.yaml
```diff
@@ -11,6 +11,9 @@ resources:
 - builder_role_binding.yaml
 - leader_election_role.yaml
 - leader_election_role_binding.yaml
+# Uncomment the following 2 lines if you want to deploy on OpenShift
```
spolti, line 14: add this information on the README so users are aware of it.

controllers/profiles/ensurer.go
```diff
@@ -24,34 +24,48 @@ import (
...
 	creator objectCreator
 }
 
+// newDummyObjectEnsurer see dummyObjectEnsurer
+func newDummyObjectEnsurer() ObjectEnsurer {
+	return &dummyObjectEnsurer{}
+}
+
+// dummyObjectEnsurer is a useful Object ensurer to apply the null pattern. Use it when you need a creator that does nothing
+type dummyObjectEnsurer struct {
```
spolti, line 53: what is the use case of it?

controllers/profiles/object_creators.go
```diff
@@ -18,7 +18,10 @@ import (
 	"context"
 	"strconv"
 
+	v1 "github.com/openshift/api/route/v1"
+
```
spolti, line 22: reorganize the imports?

README.md
```diff
@@ -69,6 +69,13 @@ make deploy IMG=<some-registry>/kogito-serverless-operator:tag
 
 This will deploy the operator into the `kogito-serverless-operator-system` namespace.
 
+**Note:** If you would like to deploy the operator OpenShift Local you need to add the `TARGET_CLUSTER_KIND` parameter
```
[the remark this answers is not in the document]
spolti: oh, I see, thanks.

controllers/profiles/ensurer.go
davidesalerno: This is used to implement the null-pattern, and so when we need an object ensurer that does nothing.
This is useful to avoid conditional code like Ricardo suggested

README.md
davidesalerno: Removed

README.md
davidesalerno: Restored the crasis

controllers/profiles/reconciler_dev.go
davidesalerno: I refactore the code based on your feedback: 
- added a PostReconcile function
- the function will be executed after the Do and won't affect the reconcile result so that we can put there operations that are not mandatory for the reconciliation like the status enrichment
- I put a default PostReconcile func into the stateSupport struct so that we can use the null-pattern: we don't need to implement it for all the states and at the same time we don't need to put conditions inside the code 

wdyt?

controllers/profiles/object_creators.go
davidesalerno: Done

operator.yaml
davidesalerno: I simplified the things based on this discussion and in particular I removed the array of `Endpoints`.
Now there is only one `Endpoint` that we are using to old the external URL the workflow will be available to.

controllers/profiles/object_creators.go
```diff
@@ -205,9 +205,32 @@ func defaultServiceCreator(workflow *operatorapi.KogitoServerlessWorkflow) (clie
...
 	}
+
 	return service, nil
 }
 
+// defaultNetworkCreator is an objectCreator for a basic Route for a workflow using dev profile
+// running on OpenShift.
+// It enables the exposition of the dev service using an OpenShift Route.
+// See: https://github.com/openshift/api/blob/d170fcdc0fa638b664e4f35f2daf753cb4afe36b/route/v1/route.crd.yaml
+func defaultNetworkCreator(workflow *operatorapi.KogitoServerlessWorkflow) (client.Object, error) {
```
spolti, line 216: Maybe it is better to move this func to the 'utils/openshift' package and rename this func to ExposeRoute or something similar.

controllers/profiles/object_creators_dev.go
```diff
@@ -0,0 +1,55 @@
...
+
+// devServiceCreator is an objectCreator for a basic Service for a workflow using dev profile
+// aiming a vanilla Kubernetes Deployment.
+// It maps the default HTTP port (80) to the target Java application webserver on port 8080.
+// It configures the Service as a NodePort type service, in this way it will be easier for a developer access the service
+func devServiceCreator(workflow *operatorapi.KogitoServerlessWorkflow) (client.Object, error) {
+	object, _ := defaultServiceCreator(workflow)
+	service := object.(*corev1.Service)
+	// Let's double-check that the workflow is using the Dev Profile we would like to expose it via NodePort
+	if IsDevProfile(workflow) {
```
spolti, line 35: can't we combine this func with devProfileServiceMutateVisitor?

controllers/profiles/profile.go
```diff
@@ -47,3 +48,12 @@ func profileBuilder(workflow *operatorapi.KogitoServerlessWorkflow) reconcilerBu
 	}
 	return profileBuilders[Profile(profile)]
 }
+
+// Function to detect if the workflow is using the Dev profile or not
+func IsDevProfile(workflow *operatorapi.KogitoServerlessWorkflow) bool {
```
spolti, line 53: maybe have a utils inside profiles to hold functions like these?

controllers/profiles/reconciler.go
```diff
@@ -76,6 +79,12 @@ func (s stateSupport) performStatusUpdate(ctx context.Context, workflow *operato
 	return true, err
 }
 
+// PostReconcile function to perform all the other operations required after the reconcile - placeholder for null pattern usages
+func (s stateSupport) PostReconcile(ctx context.Context, workflow *operatorapi.KogitoServerlessWorkflow) error {
+	//By default we don't want to perform anything after the reconcile and so we will simply return a no error
```
spolti, line 84:
```suggestion
	//By default we don't want to perform anything after the reconcile and so we will simply return no error
```

controllers/profiles/reconciler_dev.go
```diff
@@ -61,41 +65,80 @@ func (d developmentProfile) GetProfile() Profile {
 	return Development
 }
 
-func newDevProfileReconciler(client client.Client, logger *logr.Logger) ProfileReconciler {
+func newDevProfileReconciler(client client.Client, config *rest.Config, logger *logr.Logger) ProfileReconciler {
```
spolti, line 68: why call it newDevProfile, isn't better just devProfileReconciler?

controllers/profiles/reconciler_dev.go
```diff
@@ -61,41 +65,80 @@ func (d developmentProfile) GetProfile() Profile {
...
+	return &devProfileObjectEnsurers{
+		deployment:          newDefualtObjectEnsurer(support.client, support.logger, defaultDeploymentCreator),
+		service:             newDefualtObjectEnsurer(support.client, support.logger, defaultServiceCreator),
+		network:             newDefualtObjectEnsurer(support.client, support.logger, defaultNetworkCreator),
+		definitionConfigMap: newDefualtObjectEnsurer(support.client, support.logger, workflowDefConfigMapCreator),
+		propertiesConfigMap: newDefualtObjectEnsurer(support.client, support.logger, workflowDevPropsConfigMapCreator),
+	}
+}
+
+func newDevelopmentObjectEnrichers(support *stateSupport) *devProfileObjectEnrichers {
```
spolti, line 118: same

controllers/profiles/reconciler_dev.go
```diff
@@ -61,41 +65,80 @@ func (d developmentProfile) GetProfile() Profile {
...
+	return &devProfileObjectEnrichers{
+		networkInfo: newStatusEnricher(support.client, support.config, support.logger, defaultDevStatusEnricher),
+	}
+}
+
+func newDevelopmentObjectEnrichersForOpenShift(support *stateSupport) *devProfileObjectEnrichers {
```
spolti, line 124: same

controllers/profiles/reconciler_prod.go
```diff
@@ -19,6 +19,8 @@ import (
 	"fmt"
```
spolti, line 19: maybe we cab create two mor packages
profiles/{prod,dev} and another package for common functions, or just leave it on profiles package

controllers/profiles/reconciler_prod_test.go
```diff
@@ -39,15 +41,16 @@ func Test_reconcilerProdBuildConditions(t *testing.T) {
 	platform := test.GetKogitoServerlessPlatformInReadyPhase("../../config/samples/"+test.KogitoServerlessPlatformWithCacheYamlCR, t.Name())
 	client := test.NewKogitoClientBuilder().WithRuntimeObjects(workflow, platform).Build()
 
-	result, err := NewReconciler(client, &logger, workflow).Reconcile(context.TODO(), workflow)
+	config := &rest.Config{}
+	result, err := NewReconciler(client, config, &logger, workflow).Reconcile(context.TODO(), workflow)
```
spolti, line 45: maybe rename the NewReconciler to Reconciler only?

controllers/profiles/status_enricher.go
```diff
@@ -0,0 +1,60 @@
...
+		enricher: enricher,
+	}
+}
+
+// statusEnricherFn is the func that creates the initial reference object, if the object doesn't exist in the cluster, this one is created.
+// Can be used as a reference to keep the object immutable
+type statusEnricherFn func(ctx context.Context, client client.Client, config *rest.Config, workflow *operatorapi.KogitoServerlessWorkflow) (client.Object, error)
+
+// statusEnricher provides the engine for a ReconciliationState that needs to create or update a given Kubernetes object during the reconciliation cycle.
+type statusEnricher struct {
```
spolti, line 44: maybe enricherStatus?

controllers/profiles/status_enricher.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	operatorapi "github.com/kiegroup/kogito-serverless-operator/api/v1alpha08"
+)
+
+// newStatusEnricher see defaultObjectEnsurer
+func newStatusEnricher(client client.Client, config *rest.Config, logger *logr.Logger, enricher statusEnricherFn) *statusEnricher {
```
spolti, line 30: new, is this a pattern that was used?

controllers/profiles/status_enricher_dev.go
```diff
@@ -0,0 +1,92 @@
...
+	// If we aren't on OpenShift we will enrich the status with 2 info:
+	// - Address the service can be reached
+	// - Node port used
+	service := &v1.Service{}
+	err := client.Get(ctx, types.NamespacedName{Namespace: workflow.Namespace, Name: workflow.Name}, service)
+	if err != nil {
+		return nil, err
+	}
+	//If the service has got a Port that is a nodePort we have to use it to create the workflow's NodePort Endpoint
+	if service.Spec.Ports != nil && len(service.Spec.Ports) > 0 {
```
spolti, line 46: I am not sure if we really need it, what is the purpose of getting the Pod's internal IP address?
NodePort uses the node's exposed IP, no?

I mean, from outside the cluster we can't access the pod ip directly.

controllers/profiles/status_enricher_dev.go
```diff
@@ -0,0 +1,92 @@
...
+				return nil, err
+			}
+
+			labelSelector := metav1.LabelSelector{MatchLabels: labels(workflow)}
+			listOptions := metav1.ListOptions{
+				LabelSelector: klabels.Set(labelSelector.MatchLabels).String(),
+			}
+			pods, _ := clientset.CoreV1().Pods(workflow.Namespace).List(ctx, listOptions)
+			var ipaddr string
+			for _, p := range pods.Items {
```
spolti, line 59: I am not sure if we really need it, what is the purpose of getting the Pod's internal IP address?
NodePort uses the node's exposed IP, no?

I mean, from outside the cluster we can't access the pod ip directly.

controllers/profiles/status_enricher_dev.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func devStatusEnricherForOpenShift(ctx context.Context, client client.Client, config *rest.Config, workflow *operatorapi.KogitoServerlessWorkflow) (client.Object, error) {
+	// On OpenShift we need to retrieve the Route to have the URL the service is available to
+	route := &openshiftv1.Route{}
+	err := client.Get(ctx, types.NamespacedName{Namespace: workflow.Namespace, Name: workflow.Name}, route)
+	if err != nil {
+		return nil, err
+	}
+	var url *apis.URL
```
spolti, line 82: do we really need to import knative/apis here to create a http or https url?

go.mod
```diff
@@ -23,8 +23,9 @@ require (
 )
 
 require (
-	github.com/RHsyseng/operator-utils v1.4.11
+	github.com/RHsyseng/operator-utils v1.4.12
```
spolti, line 26: @davidesalerno seems that operator-utils is being used already :)

test/kubernetes_cli.go
```diff
@@ -37,6 +40,14 @@ func NewKogitoClientBuilder() *fake.ClientBuilder {
 	return fake.NewClientBuilder().WithScheme(s)
 }
 
+// NewKogitoClientBuilderWithAdditionalScheme creates a new fake.ClientBuilder with the right scheme references
+func NewKogitoClientBuilderWithAdditionalScheme(gv schema.GroupVersion, types ...runtime.Object) *fake.ClientBuilder {
```
spolti, line 44: New :)

controllers/profiles/reconciler_dev.go
davidesalerno: It is a code convention in go pre-pending the new for methods that are in charge of creating objects.
In this case we don't want to export the method outside so we are using the lower case.
I used the same style/convention because I've seen that is in line with what @ricardozanini introduced

controllers/profiles/reconciler_dev.go
```diff
@@ -61,41 +65,80 @@ func (d developmentProfile) GetProfile() Profile {
...
-		propertiesConfigMap: newObjectEnsurer(support.client, support.logger, workflowDevPropsConfigMapCreator),
+		deployment:          newDefualtObjectEnsurer(support.client, support.logger, defaultDeploymentCreator),
+		service:             newDefualtObjectEnsurer(support.client, support.logger, devServiceCreator),
+		network:             newDummyObjectEnsurer(),
+		definitionConfigMap: newDefualtObjectEnsurer(support.client, support.logger, workflowDefConfigMapCreator),
+		propertiesConfigMap: newDefualtObjectEnsurer(support.client, support.logger, workflowDevPropsConfigMapCreator),
+	}
+}
+
+func newDevelopmentObjectEnsurersForOpenShift(support *stateSupport) *devProfileObjectEnsurers {
```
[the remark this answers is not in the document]
davidesalerno: See [previous answer ](https://github.com/kiegroup/kogito-serverless-operator/pull/78#discussion_r1152960513)

controllers/profiles/reconciler_dev.go
davidesalerno: See [previous answer ](https://github.com/kiegroup/kogito-serverless-operator/pull/78#discussion_r1152960513)

controllers/profiles/reconciler_dev.go
davidesalerno: See [previous answer ](https://github.com/kiegroup/kogito-serverless-operator/pull/78#discussion_r1152960513)

go.mod
davidesalerno: Yes we introduced it after the latest release that makes it working with our kubernetes libraries ;-)

config/samples/sw.kogito_v1alpha08_kogitoserverlessworkflow_prodProfile.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: sw.kogito.kie.org/v1alpha08
+kind: KogitoServerlessWorkflow
+metadata:
+  name: greeting
+  annotations:
+    sw.kogito.kie.org/description: Greeting example on k8s!
+    sw.kogito.kie.org/version: 0.0.1
+    sw.kogito.kie.org/profile: prod
```
radtriste, line 8: Maybe adding a comment that is it not mandatory ?

controllers/profiles/status_enricher_dev.go
davidesalerno: Do you think that we can use a string?

controllers/profiles/reconciler_dev.go
```diff
@@ -61,41 +65,80 @@ func (d developmentProfile) GetProfile() Profile {
...
 	return profile
 }
 
 func newDevelopmentObjectEnsurers(support *stateSupport) *devProfileObjectEnsurers {
 	return &devProfileObjectEnsurers{
-		deployment:          newObjectEnsurer(support.client, support.logger, defaultDeploymentCreator),
-		service:             newObjectEnsurer(support.client, support.logger, defaultServiceCreator),
-		definitionConfigMap: newObjectEnsurer(support.client, support.logger, workflowDefConfigMapCreator),
-		propertiesConfigMap: newObjectEnsurer(support.client, support.logger, workflowDevPropsConfigMapCreator),
+		deployment:          newDefualtObjectEnsurer(support.client, support.logger, defaultDeploymentCreator),
```
radtriste, line 100: `newDefualt` -> `newDefault`

controllers/profiles/status_enricher.go
davidesalerno: Yes, for object creators in Go it is a de-facto convention using the New or new keyword that you can see in https://go.dev/doc/effective_go

radtriste (commented):
First comments

radtriste:
I can see from code that we have support for routes.
Could we have a simple decision logic that the operator will apply between dev/prod profile and k8s/ocp infrastructure ?
For example, what happens if I run CRC locally in prod mode ? what does it create ? nodeport or route ?

hack/go-path.sh
```diff
@@ -16,4 +16,5 @@
 # enforce GOROOT
 # shellcheck disable=SC2155
 export GOROOT=$(go env GOROOT)
-export GOPATH=$(go env GOPATH)
\ No newline at end of file
+export GOPATH=$(go env GOPATH)
+export PATH=$GOROOT:$GOPATH/bin:$PATH
```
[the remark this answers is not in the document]
davidesalerno: I did it because even if I've got the GOPATH inside my PATH, I cannot find the `goimports` binary.
Are you experimenting this issue too or is it limited to my local setup?

controllers/profiles/reconciler_prod_test.go
davidesalerno: Also for this See https://github.com/kiegroup/kogito-serverless-operator/pull/78#discussion_r1152960513

operator.yaml
```diff
@@ -469,6 +469,9 @@ spec:
               cluster:
```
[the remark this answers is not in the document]
davidesalerno: As anticipated [above](https://github.com/kiegroup/kogito-serverless-operator/pull/78#discussion_r1153055983) there is probably no need to have multiple operator.yaml at the moment and we can remove from this changes also the pre-kustomizaion stuff

hack/go-path.sh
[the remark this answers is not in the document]
radtriste: well, I have it in my .bashrc
I just don't think we should force the PATH

config/rbac/kustomization.yaml
```diff
@@ -11,6 +11,9 @@ resources:
...
+# - openshift_role.yaml
```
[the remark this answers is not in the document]
radtriste: I agree. As there is no drawbacks to do it, let's keep it simple

controllers/profiles/object_creators_dev.go
davidesalerno: We could also remove the `devProfileServiceMutateVisitor` at all

controllers/profiles/status_enricher.go
davidesalerno: I used the same approach of object creators, do you think that in this case will be better a different convention?

controllers/profiles/reconciler_dev.go
davidesalerno: I applied the null pattern as suggested ;-)

controllers/profiles/object_creators_dev.go
davidesalerno: I simplified the things as anticipated because at the moment we don't need the `devProfileServiceMutateVisitor`

controllers/profiles/profile.go
spolti: we can have a utils.go for profiles only.

controllers/profiles/reconciler_dev.go
davidesalerno: Fixed

controllers/profiles/reconciler_prod.go
davidesalerno: We can do it but since this change is not related with the main goal of this activity, isn't it better to manage in a separate task?

controllers/profiles/reconciler_dev.go
spolti: ok, thanks for the explanation.

controllers/profiles/reconciler_prod.go
spolti: Sure thing, can you please open a new jira?

controllers/profiles/status_enricher.go
spolti: nah, no need, just a matter of naming, StatusEnricher is good too.
thanks.

controllers/profiles/status_enricher_dev.go
spolti: or we can use go net library. wdyt?

spolti: sounds good @davidesalerno, thanks for the explanation.
Let's keep it.

controllers/profiles/enricher.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	operatorapi "github.com/kiegroup/kogito-serverless-operator/api/v1alpha08"
+)
+
```
desmax74, line 25: Just a suggestion, add a comment to explain the purpose of this interface for the records

controllers/profiles/reconciler_prod.go
spolti: Thanks.

spolti (approved):

main.go
```diff
@@ -105,6 +109,17 @@ func main() {
 	}
 	//+kubebuilder:scaffold:builder
 
+	// Add Route scheme if we are on OpenShift
+	if isOpenShift, err := openshift.IsOpenShift(mgr.GetConfig()); err != nil {
+		setupLog.Error(err, "Unable to verify if we are or not on OpenShift")
+		os.Exit(1)
+	} else if isOpenShift {
+		if err = routev1.AddToScheme(mgr.GetScheme()); err != nil {
```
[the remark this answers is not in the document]
davidesalerno: Yes, but once started the Operator have to be able to manage both the profiles, if we are not able to detect the cluster kind at the bootstrap in this way, there could be a serious issue and is it worth continuing in the startup?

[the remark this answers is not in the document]
desmax74: ok

config/samples/sw.kogito_v1alpha08_kogitoserverlessworkflow_withExplicitProdProfile.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: sw.kogito.kie.org/v1alpha08
+kind: KogitoServerlessWorkflow
+metadata:
+  name: greeting
+  annotations:
+    sw.kogito.kie.org/description: Greeting example on k8s!
+    sw.kogito.kie.org/version: 0.0.1
+    sw.kogito.kie.org/profile: prod #It is not mandatory to add this annotation to use the Prod Profile
```
spolti, line 8: I believe it is the default profile, right?
It would be nice to add it in the  comment as well

desmax74 (approved):

spolti (approved):

controllers/profiles/status_enricher_dev.go
```diff
@@ -0,0 +1,92 @@
...
+			clientset, err := kubernetes.NewForConfig(config)
+			if err != nil {
+				return nil, err
+			}
+
+			labelSelector := metav1.LabelSelector{MatchLabels: labels(workflow)}
+			listOptions := metav1.ListOptions{
+				LabelSelector: klabels.Set(labelSelector.MatchLabels).String(),
+			}
+			pods, _ := clientset.CoreV1().Pods(workflow.Namespace).List(ctx, listOptions)
```
MarianMacik, line 57: Why can't we use here the controller-runtime client? As far as I understand the only reason for passing `config` was to be able to create the clientset. But we can do the listing using the controller-runtime as well, can't we?

test/kubernetes_cli.go
MarianMacik, line 44: Is this used anywhere?

.github/workflows/e2e.yml
```diff
@@ -30,6 +30,8 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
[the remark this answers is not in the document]
davidesalerno: yes, unfortunately GH seems that doesn't pull the right merge commit.

controllers/profiles/status_enricher_dev.go
davidesalerno: Unfortunately we need a clientset if we would like to retrieve the external IP of the kubernetes node on which the pod is running, the controller-runtime client is not able to retrieve this information.

test/kubernetes_cli.go
davidesalerno: No, I introduced it for a test that I removed.

Do you think we can leave it there or is it better to remove this utility func as well?

.github/workflows/e2e.yml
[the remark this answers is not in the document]
MarianMacik: OK, thanks.

test/kubernetes_cli.go
MarianMacik: Sure, we can leave it here, we may use it for other tests.

config/rbac/kustomization.yaml
MarianMacik: This still holds. Is it "resolved" now and these comments are valid or are you going to create the script you mentioned?

controllers/profiles/status_enricher_dev.go
davidesalerno: Ok, I'll try even if we will need anyway the config at a devReconciler level due to [this](https://github.com/kiegroup/kogito-serverless-operator/pull/78/files#diff-cc635941ea89d2ed576d6a3dc9e4c926858209209ae85f93d3502120369c432aR108)

davidesalerno: @MarianMacik I did the change into the ensurer and tested it

controllers/profiles/ensurer.go
```diff
@@ -24,34 +24,48 @@ import (
...
+}
+
 // mutateVisitor is a visitor function that mutates the given object before performing any updates in the cluster.
 // It gets called after the objectEnforcer reference.
 //
-// The objectEnsurer will call the returned function after creating the given object structure, so callers is ensured to have the default reference of the given object.
+// The defaultObjectEnsurer will call the returned function after creating the given object structure, so callers is ensured to have the default reference of the given object.
```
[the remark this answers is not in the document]
davidesalerno: I made explicit the fact that the function is the mutateVisitor one

config/samples/sw.kogito_v1alpha08_kogitoserverlessworkflow_withExplicitProdProfile.yaml
davidesalerno: I added this detail

config/rbac/kustomization.yaml
MarianMacik, line 14: As I understand, this is not needed anymore.

davidesalerno: Done ;-)

MarianMacik (approved):

PR #87, opened 2023-04-04
[WIP] [KOGITO-8945] Defined a Traits API to customize a Kubernetes Service

amadhusu:
**Description of the change:** https://issues.redhat.com/browse/KOGITO-8945

**Motivation for the change:** Setting up Traits API similar to Camel's approach to enhance the Profile model of deployment.

**Checklist**

<details>
<summary>
</summary>

</details>


api/v1alpha08/kogitoserverlessplatform_types.go
```diff
@@ -86,8 +87,39 @@ type KogitoServerlessPlatformSpec struct {
...
+	Traits Traits `json:"traits,omitempty"`
 }
 
+type Traits struct {
+	// The configuration of Service trait
+	Service *trait.ServiceTrait `property:"service" json:"service,omitempty"`
+}
+
+// A TraitSpec contains the configuration of a trait
+// Deprecated: superceded by each Trait type, left for backward compatibility.
```
davidesalerno, line 100: I think that since this is deprecated and not used at the moment we can remove it.
The same for all the other structs that are marked as deprecated.

api/v1alpha08/kogitoserverlessplatform_types.go
```diff
@@ -86,6 +87,13 @@ type KogitoServerlessPlatformSpec struct {
 	BuildPlatform api.PlatformBuildSpec `json:"platform,omitempty"`
 	// Configuration list of configuration properties to be attached to all the Workflow built from this Platform
 	Configuration ConfigurationSpec `json:"configuration,omitempty"`
+	// Trait list of traits properties attached to all the Workflow built from this Platform
```
spolti, line 90:
```suggestion
	// Trait is list of trait properties attached to all Workflows built with Platform
```

api/v1alpha08/kogitoserverlessplatform_types.go
```diff
@@ -86,6 +87,13 @@ type KogitoServerlessPlatformSpec struct {
...
+	Traits Traits `json:"traits,omitempty"`
```
spolti, line 91: shouldn't Traits be a list of Trait?

api/v1alpha08/trait/base.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2023 Red Hat, Inc. and/or its affiliates
```
spolti, line 1: instead being inside v1alpha08, maybe move to another place, maybe api/common_objects.go or something else, wdyt?

api/v1alpha08/trait/base.go
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package trait
+
+// Base type for all traits.
+type Trait struct {
```
spolti, line 18: Where this struct is being used?

api/v1alpha08/kogitoserverlessplatform_types.go
amadhusu: Traits was used plural in the case of camel because of the number of traits they had underneath the Traits struct. Looking at it now, it makes sense to be singular so I agree Traits should be a list of Trait.

davidesalerno: We will have multiple traits soon, so it is more useful to maintain the plural (see [KOGITO-8245](https://issues.redhat.com/browse/KOGITO-8255))

api/v1alpha08/trait/base.go
amadhusu: @spolti As I am not using this struct anymore, this file base.go became unnecessary. Also I moved out the trait directory under api/trait to simplify the hierarchy. Let me know what you think.

api/v1alpha08/trait/base.go
amadhusu: struct is removed as of now.

api/v1alpha08/kogitoserverlessplatform_types.go
amadhusu: incorporated these minute changes.

api/v1alpha08/kogitoserverlessplatform_types.go
```diff
@@ -86,6 +87,13 @@ type KogitoServerlessPlatformSpec struct {
 	BuildPlatform api.PlatformBuildSpec `json:"platform,omitempty"`
 	// Configuration list of configuration properties to be attached to all the Workflow built from this Platform
 	Configuration ConfigurationSpec `json:"configuration,omitempty"`
+	// Trait is list of trait properties attached to all Workflows built with Platform
```
spolti, line 90: @davidesalerno  @amadhusu the purpose of the traits is to identify which trait needs to be configured automatically, right?

if so, don't we need to define it on the cr?

@amadhusu can you please provide a example CR as well the tests for this? If the service trait is auto == true, the service for the workflows needs to be created.

amadhusu: @spolti There is a sample CR with a basic test case for checking the workflow. I will work on the controller for the implementation of this so we can add more test cases and validate the same. I will create other examples and test cases for different Service Types as well once I am done with the Controller Implementation for the same.

PR #101, opened 2023-05-08
[KOGITO-9107] Incorporte container-builder

ricardozanini:
We are incorporating the whole `container-builder` project to make it easy to make changes to the operator until the builder takes high flights in the future.

- Moved `platform` and `builder` to within `controllers` since they are not used outside. In [KOGITO-9106](https://issues.redhat.com/browse/KOGITO-9106), I will refactor the interfaces and data structure to not expose the container-builder API to our CRs.

See: https://issues.redhat.com/browse/KOGITO-9107

To avoid a hit on our operational performance. Now all the changes can be done in one place, avoiding multiple projects unnecessarily.

<details>
<summary>
</summary>

</details>

### DIFF
--- a/.github/workflows/check-container-builder.yml
+++ b/.github/workflows/check-container-builder.yml
@@ -87,4 +87,5 @@ jobs:
           pid=$(systemctl --user show podman.service  | grep ^MainPID | sed -e 's|MainPID=||')
           sudo prlimit --pid $pid --nofile=262144:262144
           sudo prlimit --pid $$ --nofile=262144:262144
+          cd container-builder
           make ${{ matrix.container-engine }}-integration-test


desmax74 (approved):
PR ok, but we need to move the https://github.com/kiegroup/container-builder/blob/main/.github/workflows/test-go.yml to run container-builder test to run docker and podman test using the local registry to test the result of the build. In the operator the e2e test usi only the Kaniko impl

spolti (approved):

davidesalerno (approved):

container-builder/Makefile
```diff
@@ -0,0 +1,109 @@
+#BUILDTAGS ?= \
```
radtriste, line 1: do we need to move that file ?

radtriste (approved):

container-builder/Makefile
ricardozanini: It's being used in the tests.

PR #130, opened 2023-05-23
[Kogito-9207] [Operator] Make QUARKUS_KOGITO_DEVSERVICES_ENABLED and QUARKUS_DEVSERVICES_ENABLED properties immutable in devmode

desmax74:
See:https://issues.redhat.com/browse/KOGITO-9207

**Checklist**

<details>
<summary>
</summary>

</details>

### DIFF
--- a/controllers/profiles/object_creators.go
+++ b/controllers/profiles/object_creators.go
@@ -31,9 +31,11 @@ import (
 )
 
 const (
-	defaultHTTPWorkflowPort = 8080
-	defaultHTTPServicePort  = 80
-	defaultContainerName    = "workflow"
+	defaultHTTPWorkflowPortInt = 8080
+	defaultHTTPWorkflowPortStr = "8080"
+	defaultContainerName       = "workflow"
+
+	defaultHTTPServicePort = 80
 
 	applicationPropertiesFileName = "application.properties"
 
@@ -55,6 +57,8 @@ const (
 	healthStartedPeriodSeconds       = 15
 	healthStartedInitialDelaySeconds = 10
 )
+
+var defaultHTTPWorkflowPortIntStr = intstr.FromInt(defaultHTTPWorkflowPortInt)
 
 // same for now
 var defaultProdApplicationProperties = defaultDevApplicationProperties
@@ -87,14 +91,14 @@ func defaultDeploymentCreator(workflow *operatorapi.KogitoServerlessWorkflow) (c
 					Containers: []corev1.Container{{
 						Name: defaultContainerName,
 						Ports: []corev1.ContainerPort{{
-							ContainerPort: defaultHTTPWorkflowPort,
+							ContainerPort: defaultHTTPWorkflowPortInt,
 							Name:          "http",
 						}},
 						LivenessProbe: &corev1.Probe{
 							ProbeHandler: corev1.ProbeHandler{
 								HTTPGet: &corev1.HTTPGetAction{
 									Path: quarkusHealthPathLive,
-									Port: intstr.FromInt(defaultHTTPWorkflowPort),
+									Port: defaultHTTPWorkflowPortIntStr,
 								},
 							},
 							TimeoutSeconds: healthTimeoutSeconds,
@@ -103,7 +107,7 @@ func defaultDeploymentCreator(workflow *operatorapi.KogitoServerlessWorkflow) (c
 							ProbeHandler: corev1.ProbeHandler{
 								HTTPGet: &corev1.HTTPGetAction{
 									Path: quarkusHealthPathReady,
-									Port: intstr.FromInt(defaultHTTPWorkflowPort),
+									Port: defaultHTTPWorkflowPortIntStr,
 								},
 							},
 							TimeoutSeconds: healthTimeoutSeconds,
@@ -112,7 +116,7 @@ func defaultDeploymentCreator(workflow *operatorapi.KogitoServerlessWorkflow) (c
 							ProbeHandler: corev1.ProbeHandler{
 								HTTPGet: &corev1.HTTPGetAction{
 									Path: quarkusHealthPathStarted,
-									Port: intstr.FromInt(defaultHTTPWorkflowPort),
+									Port: defaultHTTPWorkflowPortIntStr,
 								},
 							},
 							InitialDelaySeconds: healthStartedInitialDelaySeconds,
@@ -178,7 +182,7 @@ func defaultServiceCreator(workflow *operatorapi.KogitoServerlessWorkflow) (clie
 			Ports: []corev1.ServicePort{{
 				Protocol:   corev1.ProtocolTCP,
 				Port:       defaultHTTPServicePort,
-				TargetPort: intstr.FromInt(defaultHTTPWorkflowPort),
+				TargetPort: defaultHTTPWorkflowPortIntStr,
 			}},
 		},
 	}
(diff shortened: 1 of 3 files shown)


controllers/profiles/object_creators.go
```diff
@@ -59,11 +58,13 @@ const (
 	healthStartedInitialDelaySeconds = 10
 )
 
-var defaultDevApplicationProperties = "quarkus.http.port=" + strconv.Itoa(defaultHTTPWorkflowPort) + "\n" +
+var defaultDevApplicationProperties = "quarkus.http.port=8080\n" +
 	"quarkus.http.host=0.0.0.0\n" +
 	// We disable the Knative health checks to not block the dev pod to run if Knative objects are not available
 	// See: https://kiegroup.github.io/kogito-docs/serverlessworkflow/latest/eventing/consume-produce-events-with-knative-eventing.html#ref-knative-eventing-add-on-source-configuration
-	"org.kie.kogito.addons.knative.health-enabled=false\n"
+	"org.kie.kogito.addons.knative.health-enabled=false\n" +
```
spolti, line 65: I don't think it will work.
These are quarkus properties expected to be set here.

If we need to set envs, we should add to the Deployment instead.
As Quarkus accepts envs and system properties, in this case here, it is better to use the correspondent system property.

I don't remember the wight of the envs and sys props on quarkus smallrye config, but IIRC, system properties has preference over envs.
It might worth to check though.

spolti (changes requested):

controllers/profiles/object_creators.go
spolti: Oh, you updated it just while I was writing this comment :)

desmax74: @spolti changed with the properties

wmedvede (commented):
I left a comment guys in case it can help

controllers/profiles/object_creators.go
```diff
@@ -59,11 +58,13 @@ const (
 	healthStartedInitialDelaySeconds = 10
 )
 
-var defaultDevApplicationProperties = "quarkus.http.port=" + strconv.Itoa(defaultHTTPWorkflowPort) + "\n" +
+var defaultDevApplicationProperties = "quarkus.http.port=8080\n" +
```
ricardozanini, line 61: Please change the `defaultHTTPWorkflowPort` to string to avoid the conversion.

desmax74: `defaultHTTPWorkflowPort` is used as int here https://github.com/kiegroup/kogito-serverless-operator/pull/130/files#diff-f6579aaafc7e439d9ba8180773a600072a37fb226296cb1ea4bb4670d10a3174R104

and as intoStr https://github.com/kiegroup/kogito-serverless-operator/pull/130/files#diff-f6579aaafc7e439d9ba8180773a600072a37fb226296cb1ea4bb4670d10a3174R111

here https://github.com/kiegroup/kogito-serverless-operator/pull/130/files#diff-f6579aaafc7e439d9ba8180773a600072a37fb226296cb1ea4bb4670d10a3174R120

and here https://github.com/kiegroup/kogito-serverless-operator/pull/130/files#diff-f6579aaafc7e439d9ba8180773a600072a37fb226296cb1ea4bb4670d10a3174R129

I'll prepare three type

ricardozanini (approved):
Thank u!

ricardozanini (approved):
Thank u!

controllers/profiles/reconciler_dev_test.go
```diff
@@ -134,7 +134,7 @@ func Test_newDevProfile(t *testing.T) {
 
 	// check if the reconciliation ensures the object correctly
 	service = test.MustGetService(t, client, workflow)
-	assert.Equal(t, int32(defaultHTTPWorkflowPort), service.Spec.Ports[0].TargetPort.IntVal)
+	assert.Equal(t, int32(defaultHTTPWorkflowPortInt), service.Spec.Ports[0].TargetPort.IntVal)
```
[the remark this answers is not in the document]
desmax74: @ricardozanini  Nope, if removed the test fails because TargetPort is a IntOrString and the IntVal return a int32

spolti (approved):
Thanks Max

@wmedvede do you want to take another look?

controllers/profiles/object_creators.go
```diff
@@ -59,11 +60,15 @@ const (
...
 
-var defaultDevApplicationProperties = "quarkus.http.port=" + strconv.Itoa(defaultHTTPWorkflowPort) + "\n" +
+var defaultHTTPWorkflowPortIntStr = intstr.FromInt(defaultHTTPWorkflowPortInt)
+
+var defaultDevApplicationProperties = "quarkus.http.port=" + defaultHTTPWorkflowPortStr + "\n" +
 	"quarkus.http.host=0.0.0.0\n" +
 	// We disable the Knative health checks to not block the dev pod to run if Knative objects are not available
 	// See: https://kiegroup.github.io/kogito-docs/serverlessworkflow/latest/eventing/consume-produce-events-with-knative-eventing.html#ref-knative-eventing-add-on-source-configuration
-	"org.kie.kogito.addons.knative.health-enabled=false\n"
+	"org.kie.kogito.addons.knative.health-enabled=false\n" +
```
wmedvede, line 69: My only comment here guys @desmax74 @spolti is that according with a chat we had yesterday with @hbelmiro , he's doing a change in the knative eventing addon, and the name of this property will now be

org.kie.kogito.addons.knative.**eventing**.health-enabled 

see that **eventing** was incorporated to the name.

Could you please confirm @hbelmiro 

Other than this, from what I can understand :) in this PR, I think it's good.

hbelmiro: Actually, the property names are not going to change. They already have that `.eventing.`, but this is not in the docs.
I opened this PR to fix it in the docs: https://github.com/kiegroup/kogito-docs/pull/353

spolti: @desmax74 can you please update the properties?

spolti: btw, ins't better to have these flags set on image side with default value if not set?

hbelmiro (approved):

controllers/profiles/object_creators.go
ricardozanini: I've fixed on my PR already. We can merge this mine will fup.

controllers/profiles/reconciler_dev_test.go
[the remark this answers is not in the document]
ricardozanini: Therefore you can define this var as int32 directly since it's meant to be used with IntOrString.

wmedvede (approved):

controllers/profiles/reconciler_dev_test.go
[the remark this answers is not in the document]
desmax74: @ricardozanini 
We can't because we can't use int32 in the IntOrStr that is used in several part of the code 

- 2 HttpGet.Port IntOrStr 
- 1 Ports.TargetPort IntOrStr 
- 1 Containers.port int32 (here is converted without problem from the int value)

 the only place where is asked an explicit int32 without automatic conversion is this test, to remove here I have to add another const with same 8080 value but of type int32.
wdyt ?

controllers/profiles/object_creators.go
desmax74: Updated

davidesalerno (approved):

PR #139, opened 2023-05-30
[KOGITO-8961][KSW-Operator] Split use case and test examples

desmax74:
See:https://issues.redhat.com/browse/KOGITO-8961

**Checklist**

<details>
<summary>
</summary>

</details>


controllers/kogitoserverlessbuild_controller_test.go
```diff
@@ -31,11 +31,12 @@ import (
 
 func TestKogitoServerlessBuildController(t *testing.T) {
 	namespace := t.Name()
-	ksw := test.GetKogitoServerlessWorkflow("../config/samples/"+test.KogitoServerlessWorkflowSampleYamlCR, namespace)
+	ksw := test.GetBaseServerlessWorkflowNoPackage(namespace)
```
ricardozanini, line 34: What package means?

test/e2e/workflow_test.go
```diff
@@ -224,7 +224,7 @@ var _ = Describe("Kogito Serverless Operator", Ordered, func() {
 
 			EventuallyWithOffset(1, func() error {
 				cmd := exec.Command("kubectl", "delete", "-f", filepath.Join(projectDir,
-					"config/samples/"+test.KogitoServerlessWorkflowSampleDevModeYamlCR), "-n", namespace)
+					"test/samples/"+test.KogitoServerlessWorkflowSampleDevModeYamlCR), "-n", namespace)
```
ricardozanini, line 227: Can u do the same here instead of using consts?

controllers/kogitoserverlessbuild_controller_test.go
desmax74: We have some test that needs as  a path "../config/samples/" to load the yaml's files, others "../../config/samples/" because they are inside a package

test/yaml.go
```diff
@@ -28,16 +28,12 @@ import (
...
-	KogitoServerlessPlatformWithCacheYamlCR                         = "sw.kogito_v1alpha08_kogitoserverlessplatform_withCacheAndCustomization.yaml"
-	KogitoServerlessPlatformWithCacheMinikubeYamlCR                 = "sw.kogito_v1alpha08_kogitoserverlessplatform_withCache_minikube.yaml"
-	KogitoServerlessPlatformYamlCR                                  = "sw.kogito_v1alpha08_kogitoserverlessplatform.yaml"
-	KogitoServerlessPlatformWithBaseImageYamlCR                     = "sw.kogito_v1alpha08_kogitoserverlessplatformWithBaseImage.yaml"
-	KogitoServerlessPlatformWithDevBaseImageYamlCR                  = "sw.kogito_v1alpha08_kogitoserverlessplatformWithDevBaseImage.yaml"
-	kogitoServerlessOperatorBuilderConfig                           = "kogito-serverless-operator-builder-config_v1_configmap.yaml"
+	KogitoServerlessWorkflowSampleYamlCR        = "sw.kogito_v1alpha08_kogitoserverlessworkflow.yaml"
+	KogitoServerlessWorkflowSampleDevModeYamlCR = "sw.kogito_v1alpha08_kogitoserverlessworkflow_devmode.yaml"
+	//@TODO can we remove ? KogitoServerlessPlatformWithCacheYamlCR                         = "sw.kogito_v1alpha08_kogitoserverlessplatform_withCacheAndCustomization.yaml"
```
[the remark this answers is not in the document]
desmax74: ok

test/e2e/workflow_test.go
desmax74: Ok, moved the constants here

config/samples/sw.kogito_v1alpha08_kogitoserverlessplatformWithBaseImage.yaml
```diff
@@ -1,10 +0,0 @@
-apiVersion: sw.kogito.kie.org/v1alpha08
```
spolti, line 1: didn't we discuss the we wouldn't change the config/samples?

desmax74: We have agreed the proposal of @davidesalerno to use the needed yaml files and create the structs with the few fields needed in the tests instead to have one yaml file for each test, this file is replaced by the platform plus the field with the base image

controllers/kogitoserverlessbuild_controller_test.go
ricardozanini: I think we should use a smarter way of doing that. Maybe we can infer the path we are and compose it dynamically. If you think it is worth a shot, it would be a great improvement. I wouldn't do it if it takes too much time.

config/samples/sw.kogito_v1alpha08_kogitoserverlessplatformWithBaseImage.yaml
ricardozanini: The `config/samples` directory is used by the OLM/Operator SDK to compose the CSV. Meaning that users will these examples on the OpenShift console. 

Bear that in mind and keep only the simplest ones. I think this is the proposal, right?

ricardozanini: So we should keep here only the ones that will be in the CSV, **one very simple example per custom resource**.

ricardozanini: For example, see here: https://operatorhub.io/operator/nexus-operator-m88i 
Click on "View Example".

That's the purpose of this directory.

desmax74: @ricardozanini  Yep the file used by kustomization 
https://github.com/kiegroup/kogito-serverless-operator/blob/main/config/samples/kustomization.yaml
are only: 
sw.kogito_v1alpha08_kogitoserverlessworkflow.yaml
sw.kogito_v1alpha08_kogitoserverlessplatform.yaml 

and in the folder in the PR we leave only these two and sw.kogito_v1alpha08_kogitoserverlessworkflow_devmodeWithConfigMapAndExternalResource.yaml with a complete example, but we can move it in the docs folder

controllers/kogitoserverlessbuild_controller_test.go
desmax74: Ok, since the performance aren't the priorities in the test I can use something to identify who is the caller and use a different path

test/yaml.go
```diff
@@ -119,3 +120,83 @@ func GetKogitoServerlessOperatorBuilderConfig(path, namespace string) *corev1.Co
 	cm.Namespace = namespace
 	return cm
 }
+
+func GetPathForSamples(path string) string {
+	operatorPath := strings.Split(path, "kogito-serverless-operator")[1]
+	packages := strings.Count(operatorPath, "/")
+	if strings.Contains(operatorPath, "/test/") || packages == 3 {
```
ricardozanini, line 127: Not required, we can do it later. But maybe the `filepath` package might help here? https://gobyexample.com/file-paths

ricardozanini (approved):

controllers/kogitoserverlessbuild_controller_test.go
desmax74: @ricardozanini  is it ok for your this https://github.com/kiegroup/kogito-serverless-operator/pull/139/files#diff-8affcd5838f425d5db07c12738684112559a18621ec539e9b33986abe804db7fR124-R132 in conjunction with the runtime.Caller ? No other ideas to decide the path, but open if you have better

ricardozanini: See my comment here: https://github.com/kiegroup/kogito-serverless-operator/pull/139#discussion_r1213170430

😆

config/samples/sw.kogito_v1alpha08_kogitoserverlessplatformWithBaseImage.yaml
spolti: I think we should keep the config/samples as it is, not remove anything, maybe the similar examples, e.g. greeting, and greeting-ITA.
it was my understand from what we discussed on Tuesday.

ricardozanini (approved):

config/samples/sw.kogito_v1alpha08_kogitoserverlessplatformWithBaseImage.yaml
spolti: AFAIK, only the examples added in the kustomize files will be added to the CSV, we can keep then around that it will not hurt anything and can serve as examples with more complex workflows.

test/yaml.go
desmax74: We do later, is a func used only in test

desmax74:
@davidesalerno  @radtriste

davidesalerno (approved):

PR #150, opened 2023-06-05
[KOGITO-9263] Add the new resources to workflow CR

ricardozanini:
In this PR we introduce a new way to specify workflow resources.

In the CRD, there's a new `Resources` attribute where one can map a local `ConfigMap` to the given workflow:

```yaml
apiVersion: sw.kogito.kie.org/v1alpha08
kind: KogitoServerlessWorkflow
metadata: 
  name: serverless-workflow-service-calls-quarkus
  annotations: 
    sw.kogito.kie.org/description: Serverless Workflow Service calls example on k8s!
    sw.kogito.kie.org/version: 0.0.1
    sw.kogito.kie.org/profile: dev
spec: 
    flow: 
       - 
    resources: 
      - configMaps:
         - configMap:
                name: openapi-configmap
            workflowPath: specs
```

The `configMaps` holds all these references. I decided to go this way 'cause in the future we can have other external sources such as `Secrets` or `Volumes`.

In _devmode_, these files can change as the developer/tooling see fit while tailoring the workflow. For builds supported by the platform, these files are mounted in the build context and added to the final immutable image as a local file system resource.

In the _worklflowproj_ module, we are generating the `ConfigMaps` grouped by **path**. By default, the `path` is `specs`. So any file added via `AddResource` will be on the same CM. One can use `AddResourceAt` instead to group the file in a different CM. See the test case. 

User Experience. See https://issues.redhat.com/browse/KOGITO-9263

**Checklist**

- [x] Testing on a cluster the devmode and ocp BC modes

<details>
<summary>
</summary>

</details>


api/v1alpha08/kogitoserverlessworkflow_types.go
```diff
@@ -35,8 +56,9 @@ type KogitoServerlessWorkflowStatus struct {
 	// +optional
 	Address duckv1.Addressable `json:"address,omitempty"`
 	// keeps track of how many failure recovers a given workflow had so far
-	RecoverFailureAttempts int       `json:"recoverFailureAttempts,omitempty"`
-	Endpoint               *apis.URL `json:"endpoint,omitempty"`
+	RecoverFailureAttempts int         `json:"recoverFailureAttempts,omitempty"`
+	LastTimeRecoverAttempt metav1.Time `json:"lastTimeRecoverAttempt,omitempty"`
```
desmax74, line 60: How can be useful the information in LastTimeRecoverAttempt ?

workflowproj/resources.go
```diff
@@ -21,38 +21,45 @@ import (
...
+	CamelRouteResource
+	GenericResource
 )
 
-// ParseResourceType tries to parse the contents of the given resource and find the correct type.
+// ParseResourceKind tries to parse the contents of the given resource and find the correct type.
 // Async and OpenAPI files are pretty fast to parse (0.00s).
 // Camel and generic files can take a fair price from the CPU (0.03s on the i5) since it takes more processing power.
-func ParseResourceType(contents string) metadata.ExtResType {
+func ParseResourceKind(contents []byte) ResourceKind {
```
desmax74, line 38: Why change the signature, the content from the configmap is returned as a string and only here https://github.com/kiegroup/kogito-serverless-operator/pull/150/files#diff-b07f106da84e1629f797884f0abd0c38ccca6628da08911e656c19e9da9edec5R42 is asked a byte[]

desmax74 (approved):

workflowproj/resources.go
ricardozanini: This is not used anymore, I just kept it here because can be useful in the future to guess the types and make a more useful/lightweight manifest for users. `[]byte` is usually returned from marshallers/conversions/read functions, that's why I've change it.

api/v1alpha08/kogitoserverlessworkflow_types.go
ricardozanini: Please take a look at the devmode reconciliation algorithm in this PR. I even wrote a comment about this.

ederign (approved):

ederign:
Tested in my local with both simple and complete project (service)  ! Thanks Zanini!

davidesalerno (approved):

spolti (approved):

controllers/profiles/reconciler_dev.go
```diff
@@ -175,7 +176,11 @@ func (e *ensureRunningDevWorkflowReconciliationState) Do(ctx context.Context, wo
 
 	externalCM, err := workflowdef.FetchExternalResourcesConfigMapsRef(e.client, workflow)
 	if err != nil {
-		e.logger.Error(err, "External Resources ConfigMap not found")
+		workflow.Status.Manager().MarkFalse(api.RunningConditionType, api.ExternalResourcesNotFoundReason, "external Resources ConfigMap not found: %s", err.Error())
```
radtriste, line 179:
```suggestion
		workflow.Status.Manager().MarkFalse(api.RunningConditionType, api.ExternalResourcesNotFoundReason, "External Resources ConfigMap not found: %s", err.Error())
```

test/e2e/workflow_test.go
```diff
@@ -213,34 +213,34 @@ var _ = Describe("Kogito Serverless Operator", Ordered, func() {
 			By("creating an instance of the Kogito Serverless Workflow in DevMode")
 			EventuallyWithOffset(1, func() error {
 				cmd := exec.Command("kubectl", "apply", "-f", filepath.Join(projectDir,
-					test.GetServerlessWorkflowE2eTest()), "-n", namespace)
+					test.GetServerlessWorkflowE2eOrderProcessingFolder()), "-n", namespace)
```
radtriste, line 216: Please change the name of the test (line #211) as you test here `order processing` and not `greeting`

test/testdata/order-processing/05_sw.kogito_v1alpha08_kogitoserverlessworkflow_devmode_orderprocessing.yaml
```diff
@@ -0,0 +1,53 @@
...
+  annotations:
+    sw.kogito.kie.org/description: Workflow for processing Orders and produce Logistics Events
+    sw.kogito.kie.org/version: 1.0.0
+    sw.kogito.kie.org/profile: dev
+spec:
+  resources:
+    configMaps:
+      - configMap:
+           name: orderprocessing-subflows
+  flow:
```
radtriste, line 28: Please add `specVersion` and `id` which are required fields in the sdk

radtriste:
```suggestion
  flow:
    id: orderprocessing
    specVersion: "0.8"
```

radtriste: https://github.com/serverlessworkflow/sdk-go/blob/main/model/workflow.go#L98
https://github.com/serverlessworkflow/sdk-go/blob/main/model/workflow.go#L71

ricardozanini: @radtriste they are required, but on our API interface, the ID is the **CR name** and the version is on the annotation. @spolti this needs to be taken care of. On our side, the API should relax these two fields.

I don't think duplicating attributes in the flow is the way to go. What I mean is, the CR name already gives an **id**. And the version of the object should be handled in the annotation.

.github/workflows/checks.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           cache: true
 
       - name: Run tests
-        run: make test
+        run: make test test-workflowproj
```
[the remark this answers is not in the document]
ricardozanini: it's another go module, that's why we have a new entry here. Go won't test it as part of the project's root via `go test`. :/

test/testdata/order-processing/05_sw.kogito_v1alpha08_kogitoserverlessworkflow_devmode_orderprocessing.yaml
spolti: The webhook will add both.
It was is already added here https://github.com/kiegroup/kogito-serverless-operator/blob/main/controllers/workflowdef/json.go#L57

@radtriste what the issue is about?

.github/workflows/checks.yml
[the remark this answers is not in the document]
radtriste: ok

ricardozanini:
Merged duo to the time we have. @radtriste if you spot anything, let me know.

PR #156, opened 2023-06-12
Fixing workflowproj module importing latest Operator apis

davidesalerno:

With this fix, everyone will be able to use the `workflowproj` go module outside the Kogito Servereless Workflow Operator

We cannot use the `workflowproj` go module outside the Kogito Servereless Workflow Operator, due to the fact that we are using a `replace` directive into the `go.mod` and at the same time the `api` go module is without version.

**Checklist**

<details>
<summary>
</summary>

</details>

### DIFF
--- a/workflowproj/go.mod
+++ b/workflowproj/go.mod
@@ -2,11 +2,8 @@ module github.com/kiegroup/kogito-serverless-operator/workflowproj
 
 go 1.19
 
-// Internal dependencies
-replace github.com/kiegroup/kogito-serverless-operator/api v0.0.0 => ../api
-
 require (
-	github.com/kiegroup/kogito-serverless-operator/api v0.0.0
+	github.com/kiegroup/kogito-serverless-operator/api v0.0.0-20230612141803-7bb1dd4abd50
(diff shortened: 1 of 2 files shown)


workflowproj/go.mod
```diff
@@ -39,6 +35,7 @@ require (
 	github.com/google/uuid v1.3.0 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
+	github.com/kiegroup/kogito-serverless-operator/api v0.0.0-20230609141400-51026c39778d // indirect
```
radtriste, line 38: that means when branching/releasing, we need to update that reference right ?

desmax74 (approved):

workflowproj/go.mod
ricardozanini: In this process, replace here with the latest tag, and then do a go mod download/tidy.

ricardozanini (approved):

spolti (approved):

radtriste (approved):

ricardozanini (approved):

PR #157, opened 2023-06-12
KOGITO-9291 Build multiplatform

radtriste:
https://issues.redhat.com/browse/KOGITO-9291

### DIFF
--- a/.ci/jenkins/Jenkinsfile.promote
+++ b/.ci/jenkins/Jenkinsfile.promote
@@ -18,8 +18,8 @@ pipeline {
         KOGITO_CI_EMAIL_TO = credentials("${JENKINS_EMAIL_CREDS_ID}")
 
         // Linked to node label
-        CONTAINER_ENGINE='podman'
-        CONTAINER_ENGINE_TLS_OPTIONS='--tls-verify=false'
+        CONTAINER_ENGINE = 'podman'
+        CONTAINER_ENGINE_TLS_OPTIONS = '--tls-verify=false'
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -155,31 +155,16 @@ podman-build: test ## Build container image with the manager.
 docker-push: ## Push docker image with the manager.
 	docker push ${IMG}
 
-# PLATFORMS defines the target platforms for  the manager image be build to provide support to multiple
```
desmax74, line 158: This was added following the operator SDK instructions, can we leave for the future upgrades of the sdk ?

.ci/jenkins/Jenkinsfile.promote
```diff
@@ -18,8 +18,8 @@ pipeline {
         KOGITO_CI_EMAIL_TO = credentials("${JENKINS_EMAIL_CREDS_ID}")
 
         // Linked to node label
-        CONTAINER_ENGINE='podman'
-        CONTAINER_ENGINE_TLS_OPTIONS='--tls-verify=false'
+        CONTAINER_ENGINE = 'podman'
```
desmax74, line 21: Why we have podman in the promote and docker in the deploy ?

radtriste: in deploy, we need `docker` for the multiplatform build
in promote, we need the a container engine only for login to registry, so `podman` should be the default

Makefile
radtriste: will set back.
Just did not want it to be confusing

ricardozanini (approved):

spolti (approved):

PR #245, opened 2023-08-25
[KOGITO-9748] Fix passing arguments to internal Kaniko builder

ricardozanini:
In this PR, we fix a problem when passing a few variables defined on `.spec.template` in `SonataFlowBuild`. Also, it was impossible to override the `QUARKUS_EXTENSIONS` env variable without `ARG` in our builder dockerfile. It was used to do an integration test with kaniko `--build-arg`.

TODO

- [ ] - Submit a doc update to teach users how to add custom extensions when building applications.

Relates to (not dependant):

- https://github.com/kiegroup/kogito-images/pull/1678

See: https://issues.redhat.com/browse/KOGITO-9748

**Checklist**

<details>
<summary>
</summary>

</details>

### DIFF
--- a/config/manager/sonataflow_builder_dockerfile.yaml
+++ b/config/manager/sonataflow_builder_dockerfile.yaml
@@ -1,5 +1,8 @@
 FROM quay.io/kiegroup/kogito-swf-builder-nightly:latest AS builder
-  
+
+ # variables that can be overridden by the builder
+ARG QUARKUS_EXTENSIONS
+
   # Copy from build context to skeleton resources project
 COPY --chmod=644 * ./resources/
 
--- a/bundle/manifests/sonataflow-operator-builder-config_v1_configmap.yaml
+++ b/bundle/manifests/sonataflow-operator-builder-config_v1_configmap.yaml
@@ -2,9 +2,10 @@ apiVersion: v1
 data:
   DEFAULT_BUILDER_RESOURCE_NAME: Dockerfile
   DEFAULT_WORKFLOW_EXTENSION: .sw.json
-  Dockerfile: "FROM quay.io/kiegroup/kogito-swf-builder-nightly:latest AS builder\n
-    \ \n  # Copy from build context to skeleton resources project\nCOPY --chmod=644
-    * ./resources/\n\nRUN /home/kogito/launch/build-app.sh ./resources\n  \n  #=============================\n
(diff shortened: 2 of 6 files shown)


config/manager/sonataflow_builder_dockerfile.yaml
```diff
@@ -1,5 +1,8 @@
 FROM quay.io/kiegroup/kogito-swf-builder-nightly:latest AS builder
-  
+
+ # variables that can be overridden by the builder
+ARG QUARKUS_EXTENSIONS
```
radtriste, line 4: `QUARKUS_EXTENSIONS` is supposed to be an env of `swf-builder` image. Why defining it as a build arg ?
https://github.com/kiegroup/kogito-images/blob/3f454b7b5f040b3e14aa4b65c8af06e3ee81cb27/modules/kogito-swf/common/scripts/module.yaml#L7

ricardozanini: Cause we can't set an env var in the build process without passing it as an `ARG` to the `Dockerfile`.

radtriste: yeah. seems that build arg definition are not possible in Cekit ...

ricardozanini: @spolti do u know why not? ^

desmax74 (approved):

config/manager/sonataflow_builder_dockerfile.yaml
radtriste: That also means we need to review all var used in those scripts: https://github.com/kiegroup/kogito-images/tree/main/modules/kogito-swf/common/scripts/added
and put them as ARG ...
Not the best

spolti: No, we might ask in the CeKit channel and propose a new feature.

spolti: my only point of observation of doing this is, at least using buildah/podman, the build will fail if the ARG in question is not set.

ricardozanini: This is meant for Kaniko/BuildConfig, and it's working ok. Can u propose on cekit to add `ARG` in the `image.yaml` structure? Maybe with a default value to avoid errors in runtime? 🤔

radtriste (approved):

PR #255, opened 2023-09-11
[KOGITO-9265] Add the PodSpecTemplate to the SonataFlow CR

ricardozanini:
In this PR:

1. Complete refactor of `controllers/profiles` method. Now more clear and easier to maintain. Each profile has its own module and a common one to share features between them. Smaller names, funcs, and revisited structs. Unused code removed.
2. Since now the `SonataFlow` CRD is huge (thanks, `podSpecTemplate`!) we needed to change from `kubectl apply` to `kubectl create` in CRDs creation. I'll try to revisit the makefile since the `deploy` and `install` are trying to apply the same artifacts. The first one should only install the CRDs, the latter the other resources.

TODO:

- [x] Reconcile the `podSpecTemplate` considering the specific user's custom template. Add guards to not override the controller's defaults like volumes, container name and etc.
- [x] Unit tests
- [x] Add a few custom attributes to the template in the e2e
- [x] Documentation - https://github.com/apache/incubator-kie-kogito-docs/pull/506

See https://issues.redhat.com/browse/KOGITO-9265

<details>
<summary>
</summary>

</details>


api/v1alpha08/sonataflow_types.go
```diff
@@ -16,14 +16,28 @@ package v1alpha08
...
 
+// FlowPodSpecTemplate describes the desired custom Kubernetes PodSpecTemplate definition for the deployed flow.
+//
+// The FlowContainer describes the container where the actual flow is running. It will override any default definitions.
+// For example, to override the image one can use `.spec.podSpecTemplate.flowContainer.image = my/image:tag`.
+type FlowPodSpecTemplate struct {
+	// FlowContainer is the Kubernetes container where the workflow application should run.
+	// One can change this attribute in order to override the defaults provided by the operator.
+	// +optional
+	FlowContainer   corev1.Container `json:"flowContainer,omitempty"`
```
masayag, line 35: When FlowContainer is specified, will the build state be skipped?

ricardozanini: Yes! See my last PR.

controllers/profiles/builder/builder.go
```diff
@@ -27,19 +27,25 @@ import (
 
 const (
 	defaultProfile = metadata.ProdProfile
+	// internal profile
+	opsProfile metadata.ProfileType = "prod_for_ops"
```
masayag, line 30: what does the `ops` prefix stand for?

api/v1alpha08/sonataflow_types.go
masayag, line 1: For the SonataFlow status, when an image is specified, the condition states the following:
```
status:
  address: {}
  conditions:
  - lastUpdateTime: "2023-09-19T08:32:17Z"
    status: Unknown
    type: Built
```
Perhaps when an image is specified the status can be set to Obsolete or Skipped to highlight no build was executed by this flow.

ricardozanini: Usually `Unknown` is the convention, but I liked `Skipped` for this use case. I'll add it. :)

controllers/profiles/builder/builder.go
masayag: naming... world's toughest problem :-)

desmax74 (approved):

api/v1alpha08/sonataflow_types.go
ricardozanini: @masayag what I did was Built = False, Reason: BuildSkipped. Can you give it a try?

api/v1alpha08/sonataflow_types.go
```diff
@@ -24,6 +24,535 @@ import (
...
+		SchedulingGates:               f.SchedulingGates,
+		ResourceClaims:                f.ResourceClaims,
+	}
+}
+
+// FlowPodSpecTemplate describes the desired custom Kubernetes PodSpecTemplate definition for the deployed flow.
+//
+// The FlowContainer describes the container where the actual flow is running. It will override any default definitions.
+// For example, to override the image one can use `.spec.podSpecTemplate.flowContainer.image = my/image:tag`.
+type FlowPodSpecTemplate struct {
```
wmedvede, line 546: Doing an analogy with the Kubernetes API, where we have corev1.PodTemplateSpec, I think this type could be named FlowPodTemplateSpec instead.

api/v1alpha08/sonataflow_types.go
```diff
@@ -115,6 +644,8 @@ type SonataFlowSpec struct {
 	// Resources workflow resources that are linked to this workflow definition.
 	// For example, a collection of OpenAPI specification files.
 	Resources WorkflowResources `json:"resources,omitempty"`
+	// PodSpecTemplate describes the deployment details of this SonataFlow instance.
+	PodSpecTemplate FlowPodSpecTemplate `json:"podSpecTemplate,omitempty"`
```
wmedvede, line 648: Following the analogy with other workload resources like Job, Deployment, etc, maybe we can name this filed just Template instead of PodSpecTemplate

api/v1alpha08/sonataflow_types.go
```diff
@@ -24,6 +24,535 @@ import (
...
+// FlowPodSpecTemplate describes the desired custom Kubernetes PodSpecTemplate definition for the deployed flow.
+//
+// The FlowContainer describes the container where the actual flow is running. It will override any default definitions.
+// For example, to override the image one can use `.spec.podSpecTemplate.flowContainer.image = my/image:tag`.
+type FlowPodSpecTemplate struct {
+	// FlowContainer is the Kubernetes container where the workflow application should run.
+	// One can change this attribute in order to override the defaults provided by the operator.
+	// +optional
+	FlowContainer   FlowContainer `json:"flowContainer,omitempty"`
+	PodSpecTemplate FlowPodSpec   `json:",omitempty"`
```
wmedvede, line 551: Following analogy with the Kubernetes API -> PodTemplateSpec type, I think this filed "PodSpecTemplate" can be named just "PodSpec"

wmedvede: Silly question here to double check if my understanding is correct, so FlowContainer FlowContainer can be eventually fine tuned by the users at the time they define their deployment. And PodSpecTemplate FlowPodSpec is for internal use?

api/v1alpha08/sonataflow_types.go
```diff
@@ -24,6 +24,535 @@ import (
...
+const DefaultContainerName = "workflow"
+
+// FlowContainer is the container for the internal workflow deployment based on the default Kubernetes Container API
+type FlowContainer struct {
+	// Container image name.
+	// More info: https://kubernetes.io/docs/concepts/containers/images
+	// This field is optional to allow higher level config management to default or override
+	// container images in workload controllers like Deployments and StatefulSets.
+	// +optional
+	Image string `json:"image,omitempty" protobuf:"bytes,2,opt,name=image"`
```
wmedvede, line 36: Another silly question. So, if the user configures an image here, the operator will basically skip the image generation for that workflow, and just grab the configured one, right?
Which use case are we trying to cover with this? 
I mean, looks like a very advanced situation where the build was produced completely outside the operator scope, no warranty that the image even corresponds to a workflow, so this why I'm asking, I which situations people might want to do that? optimizations maybe?

test/testdata/sonataflow.org_v1alpha08_sonataflow-simpleops.yaml
```diff
@@ -0,0 +1,33 @@
...
+kind: SonataFlow
+metadata:
+  name: simple
+  annotations:
+    sonataflow.org/description: Simple example on k8s!
+    sonataflow.org/version: 0.0.1
+spec:
+  podSpecTemplate:
+    flowContainer:
+      image: quay.io/kiegroup/sonataflow-minimal-example:latest
```
wmedvede, line 25: To double check with my previous question, so here what we are doing is:

1) In the flow, we are defining the HelloWorld workflow.

2) then, by, an external media we have already prepared an image that has exactly the build of that workflow. In the tests we have done all this manually by doing docker build -t quay.io/kiegroup/sonataflow-minimal-example:latest /tmp/simpleworkflow/ etc. So, we know by construction that the workflow is the same, that we have quarkus project build that can start, etc.

3) So in the end, we have an image that was produced by the quay.io/kiegroup/kogito-swf-builder-nightly:latest, pretty much the same that would have been done if no image was configured.

Finally, all scenarios where the workflow defined in the "flow" is different than the "corresponding" workflow already packaged/build in the provided image are not good. why? because, the semantic of what you see in the "flow" is different from what will be executing.

Maybe I'm missing something, but I see this very risky, and other than a kind of optimization I don't see the use cases for this. ¿?

wmedvede (approved):
I have left some comments in case it could make sense and some silly questions.
Thanks.

api/v1alpha08/sonataflow_types.go
ricardozanini: Hey @wmedvede thanks for the question. Please see [https://issues.redhat.com/browse/KOGITO-9833](KOGITO-9833) where I summarize the fix for this problem. This is the first question one will have once it tries this feature. The operator uses the `spec.flow` to do any configurations like discovery and eventing objects creation. So if one sets an image that is not compatible, it happens to have a bad deployment in the end. But we will do an integrity check before deploying it in the future.

api/v1alpha08/sonataflow_types.go
ricardozanini: No, the `PodSpecTemplate` has an array of `containers` mapped by the container name. We set the application container to `workflow`. So if a user overrides the defaults, they should to change `spec.podSpecTemplate.containers[workflow].` That's cumbersome so we added a shortcut to `spec.flowContainer`. I'll clarify this in the OpenAPI + docs.

test/testdata/sonataflow.org_v1alpha08_sonataflow-simpleops.yaml
ricardozanini: > Maybe I'm missing something, but I see this very risky, and other than a kind of optimization I don't see the use cases for this. ¿?

@wmedvede I believe I replied to your questions in the other comment. As far from this one, we can't control the build process. Users might have really strict and particular requirements when building images such as hardening, security issues, company policies, and so on. Opening an API to enable custom builds in the operator can be quite challenging and it's not our core. The build system today is simple enough for really small/simple use cases, but we can't scale. For this reason, we explore the `PodSpec` for users to set their image built externally.

As I said, we will add an integrity check to the image as we progress avoiding this kind of disruption.

utils/kubernetes/volumes_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	AddOrReplaceVolume(&podSpec, volumes...)
+
+	assert.Len(t, podSpec.Volumes, 3)
+	sort.Slice(podSpec.Volumes, func(i, j int) bool {
+		return podSpec.Volumes[i].Name < podSpec.Volumes[j].Name
+	})
+	assert.Equal(t, "cmA", podSpec.Volumes[0].ConfigMap.Name)
+	assert.Equal(t, "cmB", podSpec.Volumes[1].ConfigMap.Name)
+	assert.Equal(t, "cmC", podSpec.Volumes[2].ConfigMap.Name)
```
spolti, line 54: if my understanding is correct, the volumes provided in the podSpec will be removed?

ricardozanini: no! we merge user's volumes with the operator's

api/v1alpha08/sonataflow_types.go
wmedvede: ok, thanks for the explanation.

utils/kubernetes/volumes_test.go
spolti: shouldn't we have 5 volume in this test case then?

ricardozanini: No, volume1 and volume2 are in both containers.

krisv (approved):

spolti (approved):

PR #262, opened 2023-09-24
KOGITO-9739: Create OpenShift routes with HTTPS for SonataFlows

wmedvede:

**Checklist**

- [X] Add or Modify a unit test for your change
- [X] Have you verified that tall the tests are passing?

<details>
<summary>
</summary>

</details>

### DIFF
--- a/utils/openshift/route_test.go
+++ b/utils/openshift/route_test.go
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//      http://www.apache.org/licenses/LICENSE-2.0
+//     http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


ricardozanini (approved):

spolti (commented):
For edge termination it is expected that the user provide the certificates, is there a doc about it on kogito-docs?

If so, I think it is better turn the termination configurable, so use can choose to bundle the certificates within the application and use passthrough termination instead.

Hardcoding to edge might not be good.

ricardozanini:
@spolti this is devmode, but doesn't the TLS termination edge use the OpenShift CA by default? Not sure if we need to externalize this configuration then. Additionally, we do not reconcile this route AFAIK, so a user can freely edit it if they need.

ricardozanini:
@spolti can you clarify? why won't work? The CRC/OpenShift route creation via `TLS: edge` termination works out of the box without providing custom certs. If one needs it, they can edit the route after.

krisv (approved):

ricardozanini:
> You don't need to provide custom certs, but if you use the ocp one, you will have to use insecure ssl due invalid certificates or, export client.key unless the certificate is valid and it is already in the cacerts chain.
> 
> Isn't better to keep plain text? Or is there a hard requirement for it?

@caponetto needs the route to be secured. I guess it's on their requirements to also have `--insecure` set. Although I think only CRC requires that. The default TLS on OpenShift is not self-signed, but based on the user's installation AFAIK 🤔

spolti:
+1 of having the TLS configuration available, but, it would be better to have a flag on the sonataflow spec to enable it by user decision, wdyt?

ricardozanini:
> +1 of having the TLS configuration available, but, it would be better to have a flag on the sonataflow spec to enable it by user decision, wdyt?

We can if users keep asking for it. I'd rather refrain from adding those attributes beforehand. Like I said, the route is not in the reconciliation path, so one can freely do whatever after it's created.

ricardozanini:
@wmedvede just to make sure, can you do a test for us?
1. deploy a devmode workflow on OpenShift
2. let the operator create the route
3. remove "tls: edge" and see if the operator changes it back to having this config. if so, it should not.

In summary, we should only create the route and not reconcile it, so users can change it as they please.

wmedvede:
@ricardozanini Below goes the execution sequence, and I confirm that the route is not reconciled. 

Regarding the exposure of more configuration options to the user, and let it provide certificates, etc., I think that this can be cool to do, but we must be sure in which use case this is needed (specially for dev mode). And, If there are real needs/priorities, we can do. Otherwise, the "edge" termination is fine in my opinion. And I confirm that the "default" certificate is used when no other one is provided for the "edge" termination configuration, as it's stated in the documentation.

**Route that is being created now**
RouteCreated1

RouteCreated2

**Accessing the route: (certificate in use)**

RouteAccess1

**Accessing the route: (certificate in use)**

RouteAccess2

**Accessing the route: (route is working as expected)**

RouteAccess3

**Route was manually changed, for instance to http:**

RouteChanged


**After some time, the route remains with the last change, so no reconciliation.** 

**The TLS route no longer works as expected.**

HttpsRouteNoLongerWorks

**Now the modified route works as expected:**

HttpRouteWorks

I conclude it's working, which I was already tested before sending the PR. Let's see if this fulfills the requester needs.

spolti (approved):

ricardozanini:
Thank you @wmedvede!

PR #272, opened 2023-10-11
[KOGITO-9890] Generate safe volume naming based on DNS 1035

ricardozanini:
In this PR, we safely generate volume naming for the dev profile resources to avoid problems with mount paths like `META-INF/resources`.

See https://issues.redhat.com/browse/KOGITO-9890

**Checklist**

<details>
<summary>
</summary>

</details>

### DIFF
--- a/controllers/profiles/dev/object_creators_dev.go
+++ b/controllers/profiles/dev/object_creators_dev.go
@@ -26,7 +26,6 @@ import (
 	"github.com/kiegroup/kogito-serverless-operator/controllers/profiles"
 	"github.com/kiegroup/kogito-serverless-operator/controllers/profiles/common"
 	"github.com/kiegroup/kogito-serverless-operator/controllers/workflowdef"
-	"github.com/kiegroup/kogito-serverless-operator/utils"
 	kubeutil "github.com/kiegroup/kogito-serverless-operator/utils/kubernetes"
 	"github.com/kiegroup/kogito-serverless-operator/workflowproj"
 )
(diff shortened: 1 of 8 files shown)


controllers/profiles/dev/object_creators_dev.go
```diff
@@ -139,7 +138,7 @@ func mountDevConfigMapsMutateVisitor(flowDefCM, propsCM *corev1.ConfigMap, workf
 				}
 				// the resource configMap needs a specific dir, inside the src/main/resources
 				// to avoid clashing with other configMaps trying to mount on the same dir, we create one projected per path
-				volumeMountName := configMapExternalResourcesVolumeNamePrefix + utils.PathToString(workflowResCM.WorkflowPath)
+				volumeMountName := kubeutil.MustSafeDNS1035(configMapExternalResourcesVolumeNamePrefix, workflowResCM.WorkflowPath)
```
wmedvede, line 141: sorry for the silly question, mostly becasue I'm not yet fully familiar with go..., but in the case of a wrong name, the exectuion of MustSafeDNS1035 might lead in a panic(xxxx) invocation. In that case, what happens with all this execution path? I mean, when the panic exectuion starts to "kill" the exectuion chain, what are is being killed at head of that chain. A reconcile thread? the controller?

wmedvede (commented):
LGTM, only a silly question .

controllers/profiles/dev/object_creators_dev.go
ricardozanini: It will kill the pod.

krisv (approved):

PR #283, opened 2023-10-17
Fix #304 - Operator driven service discovery API

### LINKED ISSUE #304: Operator driven service discovery API
Implement the go lang based service discovery API to resolve k8s object addresses according with https://sonataflow.org/serverlessworkflow/main/cloud/quarkus/kubernetes-service-discovery.html

wmedvede:

**Checklist**

<details>
<summary>
</summary>

</details>


ricardozanini (commented):
I like this initial (big) step. Therefore, can we please have one package? It's uncommon in golang to have that many packages for this implementation. Especially in the scope of `api` package. This is a convention for the CRD types, might be confusing to introduce this here.

What I recommend is to create one single `controllers/discovery` package and in there, we create our files with meaningful prefixes such as `parser_types.go`.  Also, we can the implementation by adopting lowercase letters in the func/types and exposing just the discovery engine like: `discovery.NewHandler(workflow).Do()`. Just an example. 

The other modules/packages don't care about unexposed data structures that we need to do the discovery.

wmedvede:
> I like this initial (big) step. Therefore, can we please have one package? It's uncommon in golang to have that many packages for this implementation. Especially in the scope of `api` package. This is a convention for the CRD types, might be confusing to introduce this here.
> 
> What I recommend is to create one single `controllers/discovery` package and in there, we create our files with meaningful prefixes such as `parser_types.go`. Also, we can implement it by adopting lowercase letters in the func/types and exposing just the discovery engine like: `discovery.NewHandler(workflow).Do()`. Just an example.
> 
> The other modules/packages don't care about unexposed data structures that we need to do the discovery.

Good suggestions, I'll apply all. Many thanks Ricardo.

controllers/discovery/uri_utils.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+	}
+}
+
+func buildURI(scheme string, host string, port int) string {
+	return fmt.Sprintf("%s://%s:%v", scheme, host, port)
+}
+
+func buildKubernetesServiceDNSUri(scheme string, namespace string, name string, port int) string {
+	return fmt.Sprintf("%s://%s.%s.svc.cluster.local:%v", scheme, name, namespace, port)
```
ricardozanini, line 88: you can remove the `.cluster.local`. It might not work with headless services. Just the Name and Namespace should resolve correctly.

controllers/discovery/uri_utils.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func buildKubernetesPodDNSUri(scheme string, namespace string, podIP string, port int) string {
+	hyphenedIp := strings.Replace(podIP, ".", "-", -1)
+	return fmt.Sprintf("%s://%s.%s.pod.cluster.local:%v", scheme, hyphenedIp, namespace, port)
```
ricardozanini, line 93: Same thing here. We can stop at `pod`. Do a few tests on your end, I think it will work fine.

controllers/discovery/kubernetes_catalog.go
```diff
@@ -0,0 +1,65 @@
...
+			} else {
+				return serviceUri, nil
+			}
+		} else {
+			if podUri, err3 := resolvePodUri(pod, "", uri.GetCustomPort(), outputFormat); err3 != nil {
+				return "", err3
+			} else {
+				return podUri, nil
+			}
+		}
```
ricardozanini, line 63:
```suggestion
			if serviceUri, err := resolveServiceUri(*service, uri.GetCustomPort(), outputFormat); err != nil {
				return "", err
			} else {
				return serviceUri, nil
			}
		} else {
			if podUri, err := resolvePodUri(pod, "", uri.GetCustomPort(), outputFormat); err != nil {
				return "", err
			} else {
				return podUri, nil
			}
		}
```
Don't need to create new vars in this case.

ricardozanini (commented):
Great work! I really like it! Just a few points if you don't mind. I will refrain from adding more since it's a draft yet. 😅

controllers/discovery/queries.go
```diff
@@ -0,0 +1,93 @@
...
+// findIngress finds an ingress by name in the given namespace.
+func findIngress(ctx context.Context, cli client.Client, namespace string, name string) (*networkingV1.Ingress, error) {
+	ingress := &networkingV1.Ingress{}
+	if err := cli.Get(ctx, buildObjectKey(namespace, name), ingress); err != nil {
+		return nil, err
+	}
+	return ingress, nil
+}
+
+func buildObjectKey(namespace string, name string) client.ObjectKey {
```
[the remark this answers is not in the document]
wmedvede: Not sure if we can here, becasue we don't have the Object, instead I have namespace and name directly.

controllers/discovery/port_utils.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func isServicePortSecure(servicePort corev1.ServicePort) bool {
+	return servicePort.Name == httpsProtocol || isSecurePort(int(servicePort.Port))
+}
+
+// findContainerPort returns the best suited PortPort to connect to a pod, or nil if the pod has no ports at all.
+// The optional customPort can be used to determine which port should be used for the communication, when not set,
+// the best suited port is returned. For this last, a secure port has precedence over a non-secure port.
+func findContainerPort(containerPorts []corev1.ContainerPort, customPort string) *corev1.ContainerPort {
```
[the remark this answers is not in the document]
wmedvede: yes, this why this method has **customPort string** as input parameter.

In the ResourceUri we have https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/df49ec869c47b2da6773cdfbf2feb1bf89057d9a/controllers/discovery/discovery.go#L16 

Ths CustomPortLabel that can be set as part of ResourceUri to query. When set, it is properly propagated and passed to this method, and has priority, so we try to get the port with that label.

The other criteria to consider ssl port, or http, or web, or the first available if no custom, ssl, http or web are present is borrowed from the kogio runtimes kubernetes catalog implementation to keep semantic compatibility.

controllers/discovery/kubernetes_catalog.go
wmedvede: I'm trying to evict the "err" variable shadowing...., while the shadowing will work, IDK if this contribute to increase compiler warnings, etc.

controllers/discovery/uri_parser.go
```diff
@@ -0,0 +1,134 @@
...
+import (
+	"fmt"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"regexp"
+	"strings"
+)
+
+const (
+	// valid namespace, name, or label name.
+	dns1123LabelFmt         string = "[a-z0-9]([-a-z0-9]*[a-z0-9])?"
```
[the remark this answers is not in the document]
wmedvede: I couldn't find this const public there, nor in the "validation" package, we have methods to validate, but not te expression.

controllers/discovery/port_utils.go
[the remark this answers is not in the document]
ricardozanini: Great, instead of `custom-port`, can we call it just `port`? It's easier to remember.

controllers/discovery/kubernetes_catalog.go
ricardozanini: This should be fine, usually we use another var name when the other vars are meant to be used in the context.

controllers/discovery/queries.go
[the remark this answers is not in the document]
ricardozanini: Ok, so then just do `types.NamespacedKey{namespace, name}` instead of creating a custom function that use an alias: https://github.com/kubernetes/apimachinery/blob/master/pkg/types/namespacedname.go#L27

wmedvede:
@ricardozanini Service discovery was integrated according to the microprofile catalog properties strategy and working for kubernetes services and pods. The other kubernetes objects I think we can add in a separate PR.
I'll take a look at the builder image in the mean time to incorporate the kogito-addons-quarkus-microprofile-config-service-catalog

ricardozanini (approved):

ricardozanini:
@wmedvede 
> The other kubernetes objects I think we can add in a separate PR

Do we have an issue for this one?

wmedvede:
> @wmedvede
> 
> > The other kubernetes objects I think we can add in a separate PR
> 
> Do we have an issue for this one?

No. but can I can open, the only are the commit comments reffering to  KOGITO-9867...., IDK, if you don't mind loosing the history of your review I can open a new PR and use an issue, otherwise I can open an issue and at least link it to this PR.

wmedvede:
Closes apache/incubator-kie-issues/issues#692

utils/kubernetes/deployment.go
```diff
@@ -111,6 +111,21 @@ func GetContainerByName(name string, podSpec *v1.PodSpec) (*v1.Container, int) {
 	return nil, -1
 }
 
+// GetContainerPortByName returns a pointer to the ContainerPort within the given Container.
+// If none found, returns nil.
+// It also returns the position where the container port was found, -1 if none.
+func GetContainerPortByName(name string, container *v1.Container) (*v1.ContainerPort, int) {
+	if container == nil {
```
tchughesiv, line 118: should this instead be `if container != nil {` and just `return nil, -1` once at the end of the function?

utils/kubernetes/service.go
```diff
@@ -42,3 +42,18 @@ func RetrieveServiceURL(service *v1.Service) (*apis.URL, error) {
 		Path:   service.Name}
 	return apis.ParseURL(url.String())
 }
+
+// GetServicePortByName returns a pointer to the ServicePort within the given Service.
+// If none found, returns nil.
+// It also returns the position where the service port por was found, -1 if none.
+func GetServicePortByName(name string, service *v1.Service) (*v1.ServicePort, int) {
+	if service == nil {
```
tchughesiv, line 50: should this instead be `if service != nil {` and just `return nil, -1` once at the end of the function?

tchughesiv (commented):
a couple nits, otherwise lgtm

utils/kubernetes/deployment.go
wmedvede: I think we have here different programming style, and really no criteria used around the project in general.
Let me explain, the intention here is that you don't have to wait to the end of the function to know what happens, instead, as soon you know the container is null, you don't need to read the remainder of the code, you know directly what the result should be.  As said, it correspond to a programming styling, all variants are good.
For this particular file, we had already a function GetContainerByName https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/1eb14dfcd37705ebcb7c60342063bf3f438532a0/utils/kubernetes/deployment.go#L102, that uses this style.
I you don't mind, I'd prefer to follow same patter for a similar function.

utils/kubernetes/service.go
wmedvede: More or less the same here, same kind of function tries to follow same pattern.

utils/kubernetes/deployment.go
tchughesiv: definitely understand the approach, but i should have noticed this pattern being used in the existing functions... my bad

wmedvede:
> @wmedvede looks good but I really miss unit tests for all the different discovery functions. There are numerous branches in each, for example we have `uri_parser_test.go` Could we also do this for other parts?

tests where added, this is good enough for now.

utils/kubernetes/deployment.go
wmedvede: yeah, we don't have really a standardized adoption, but yeah, in this case ha had this situation.

domhanak (approved):

PR #299, opened 2023-11-08
[KOGITO-9924] use AppPropertyHandler for Data Index custom properties

tchughesiv:
Leverage the `AppPropertyHandler` for Data Index custom properties.

The operator should support custom application properties for Data Index, just as it does for workflow.

**Checklist**

<details>
<summary>
</summary>

</details>

### DIFF
--- a/controllers/profiles/common/app_properties.go
+++ b/controllers/profiles/common/app_properties.go
@@ -30,6 +30,7 @@ const (
 	kogitoServiceUrlProperty         = "kogito.service.url"
 	kogitoServiceUrlProtocol         = "http"
 	dataIndexServiceUrlProperty      = "mp.messaging.outgoing.kogito-processinstances-events.url"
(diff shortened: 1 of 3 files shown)


controllers/profiles/common/app_properties.go
```diff
@@ -50,6 +51,7 @@ var immutableApplicationProperties = "quarkus.http.port=" + DefaultHTTPWorkflowP
 var _ AppPropertyHandler = &appPropertyHandler{}
 
 type AppPropertyHandler interface {
+	asService() AppPropertyHandler
```
ricardozanini, line 54: You don't need to add a private method to the interface.

ricardozanini (approved):
Just one small comment! Great work! :)

controllers/profiles/common/app_properties.go
tchughesiv: good catch, i had originally made it public

tchughesiv: i've decided to do away with that method entirely... seems cleaner.

domhanak (approved):

PR #309, opened 2023-11-22
kie-issues-308: Operator driven service discovery API Phase2

wmedvede:
- Add the discovery of kubernetes Deployment, StatefulSet, and Ingress objects to the service discovery API.

**Checklist**

<details>
<summary>
</summary>

</details>

### DIFF
--- a/controllers/discovery/discovery.go
+++ b/controllers/discovery/discovery.go
@@ -32,6 +32,10 @@ const (
 	KubernetesScheme = "kubernetes"
 	OpenshiftScheme  = "openshift"
 
+	// PortQueryParam well known query param to select a particular target port, for example when a service is being
+	// discovered and there are many ports to select.
+	PortQueryParam = "port"
+
 	// KubernetesDNSAddress use this output format with kubernetes services and pods to resolve to the corresponding
 	// kubernetes DNS name. see: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/
 	KubernetesDNSAddress = "KubernetesDNSAddress"
@@ -55,12 +59,11 @@ const (
 )
 
 type ResourceUri struct {
-	Scheme       string
-	GVK          v1.GroupVersionKind
-	Namespace    string
-	Name         string
-	Port         string
-	CustomLabels map[string]string
+	Scheme      string
+	GVK         v1.GroupVersionKind
+	Namespace   string
+	Name        string
+	QueryParams map[string]string
 }
 
 // ServiceCatalog is the entry point to resolve resource addresses given a ResourceUri.
@@ -108,9 +111,9 @@ type ResourceUriBuilder struct {
 func NewResourceUriBuilder(scheme string) ResourceUriBuilder {
 	return ResourceUriBuilder{
 		uri: &ResourceUri{
-			Scheme:       scheme,
-			GVK:          v1.GroupVersionKind{},
-			CustomLabels: map[string]string{},
+			Scheme:      scheme,
+			GVK:         v1.GroupVersionKind{},
+			QueryParams: map[string]string{},
 		},
 	}
 }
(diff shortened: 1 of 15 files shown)


wmedvede:
Closes: apache/incubator-kie-kogito-serverless-operator#308

controllers/discovery/discovery.go
```diff
@@ -57,6 +59,7 @@ type ResourceUri struct {
 	GVK          v1.GroupVersionKind
 	Namespace    string
 	Name         string
+	Port         string
```
ricardozanini, line 62: Is this only valid for Services and Containers? We might have to add this to a specific struct or have a custom query attribute here.

controllers/discovery/kubernetes_catalog.go
```diff
@@ -58,21 +72,97 @@ func (c k8sServiceCatalog) resolveServiceQuery(ctx context.Context, uri Resource
...
+		if serviceList != nil && len(serviceList.Items) > 0 {
+			referenceService := selectBestSuitedServiceByCustomLabels(serviceList, uri.CustomLabels)
+			return resolveServiceUri(referenceService, uri.GetPort(), outputFormat)
+		} else {
+			return resolvePodUri(pod, "", uri.GetPort(), outputFormat)
+		}
+	}
+}
+
+func (c k8sServiceCatalog) resolveDeploymentQuery(ctx context.Context, uri ResourceUri, outputFormat string) (string, error) {
```
ricardozanini, line 87: So, if you can't find the service, you try to find the pod assuming that there's only one pod in the deployment as a fallback? I think it's safer to just fail and not assume that. Otherwise, we might see weird behaviors in the service connectivity. Also, we can add this to the docs saying that a Deployment must have a Service.

ricardozanini (commented):
Great work man! Just a few observations if you don't mind. 👍

controllers/discovery/kubernetes_catalog.go
wmedvede: We can change if we believe that we have a better option, no problem, but let's double check.

Let me explain the calculation:

1) we have a service, all good, we use it.

2) No service is defined? 
Ok, we try to pick a pod, but not assuming there is one, we do several checks instead, look:

No replicas ? deployment.Spec.Replicas == nil || *deployment.Spec.Replicas == 0 
we return a proper message.

Too many replicas:  *deployment.Spec.Replicas > 1
we return a proper message.

Only one replica?
good!, we first try to find the corresponding ReplicaSet object, and if found, we double check again how many pods exists, if it has 0, 1, or > 1 pods.

If no pods, or too many pods, we return a proper message.

Only if one pod exists, then we return the corresponding pod address.

The criteria was borrowed from this diagram:

sw-discovery-flow



So we basically do several checks, and only returns a pod if only one pod is present.
I think it's a matter of how we define the semantic of the service discovery, and here is where we can revisit it.

From the service discovery perspective, what does it mean to access a Deployment?

**Proposal 1)** is what we have, i.e., access to it with the Service is resent, or with the Pod, only if we have one pod etc.
Does it make sense?
**Pros:** A Deployment object might exists without an associated object, so from this point of view, it make sense.

**Cons:** It can be argued that the solution is kind of limited, since only deployments with only one pod can be accessed with the service discovery if no service is present. Additionally, a Pod can go down at any time.

**Proposal 2)** we look for the Service, if we find it, all good, if not, ok, we throw an error saying that no Service was found. This basically changes a bit the semantic.
**Pros:** with the service, the associated pods can go up and down at any time, the service always resolves how to pick one for a given request.

**Cons:** Well, if no service, no access. But it can be fair assumption considering that having service is a recommended practice.

If we go with Proposal 2), we might probably need to apply same criteria for the StatefulSet?, or, in the case of the stateful set, in the case of only 1 replica, we might return **always** the corresponding dns name of the corresponding pod (no chance to return the IP), something like this:  **stateful-set-example-pod-0.stateful-set-example.test-discovery.svc.cluster.local:8080**
In the end, this DNS name, in cases we have 1 pod, will never change, so, if pod goes down at any time, when restarted, the name DNS name will resolve the instance name.

wmedvede: After re-reading... myself :) , yeah, we should go with Proposal2, at least for the Deployment. Since the ReplicaSet associated with the deployment don't warrant that only one pod instance will be created even in the case where you set the replicas == 1, and we might face situations where 2 pods can be around (at least for a period of time).

wdyt?

controllers/discovery/discovery.go
wmedvede: +1, I will go back to the custom query attribute, which I had in the previous PR.

controllers/discovery/kubernetes_catalog.go
ricardozanini: Yup, +1 for proposal 2. I think this diagram would make sense for runtimes, but not at the operator level. We can even review this diagram and add the constraint there since the guide should be relocated to Cloud > Operator.

Anyhow, it's too complex to try to "guess" if a deployment has a pod available or not, too many variables that can vary (😢 ), no service, no discovery. Same for the StatefulSet. 👍

wmedvede:
@ricardozanini suggestions were added.

ricardozanini (approved):

jakubschwan (approved):

PR #311, opened 2023-11-28
[KOGITO-9940] Add Job Service to SonataFlowPlatform CRD

jordigilh:
* Adds a new field to the `ServiceSpec` in the SonataflowPlatform CRD for a Job Service, reusing the same type as the `DataIndex` field. 
*  Implements the reconciliation logic for the Job Service.
*  Adds 2 new unit tests specific for the Job Service and another one for the use case where both DataIndex and Job Service fields are populated with an empty struct.
* Refactor platform services to split from `profiles` and move it inside `platform` package

@tchughesiv @ricardozanini PTAL.


ricardozanini (commented):
Nice work! Many thanks! I just left a few small comments.

controllers/profiles/common/app_properties.go
```diff
@@ -45,27 +45,59 @@ const (
...
+		return DataIndexServiceName
+	case JobService:
+		return JobServiceName
+	}
+	return ""
+}
+
+// GetServiceImageName returns the image name of the service's container. It takes in the service and persistence types and returns a string
+// that contains the FQDN of the image, including the tag.
+func GetServiceImageName(persistenceName string, stype int) string {
```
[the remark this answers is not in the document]
jordigilh: Forgot to add the type. I'll push a new commit with changes from your comments and include the new type.

controllers/profiles/common/app_properties.go
```diff
@@ -45,27 +45,59 @@ const (
...
 
-var discoveryLikePropertyExpr = regexp.MustCompile(discoveryLikePropertyPattern)
+const (
+	DataIndexService = iota
+	JobService
+)
 
-var _ AppPropertyHandler = &appPropertyHandler{}
+// GetContainerName returns the name of the service's container in the deployment.
+func GetContainerName(stype int) string {
```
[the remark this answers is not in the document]
jordigilh: It's used in the platform package, so it needs to be public https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/cba7063ca24d0350252b41e1de0f3ea0dfe2b0b8/controllers/platform/services.go#L186

controllers/profiles/common/app_properties.go
```diff
@@ -45,27 +45,59 @@ const (
...
-	DataIndexName      = "data-index-service"
-
-	PersistenceTypeEphemeral   = "ephemeral"
-	PersistenceTypePostgressql = "postgresql"
-
+	imageNamePrefix                          = "quay.io/kiegroup/kogito"
+	DataIndexServiceName                     = "data-index-service"
+	DataIndexName                            = "data-index"
+	JobServiceName                           = "jobs-service"
+	ImageTag                                 = "1.44"
```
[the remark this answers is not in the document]
jordigilh: Done.

controllers/profiles/common/app_properties.go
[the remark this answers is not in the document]
jordigilh: Added enum type for clarity.

controllers/profiles/common/app_properties.go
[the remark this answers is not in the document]
jordigilh: Done.

ricardozanini (approved):
Great work! ❤️

ricardozanini:
Can you run:

```
make generate-all
make vet fmt
```

Should pass the CI.

tchughesiv:
lgtm

tchughesiv (changes requested):
missed this earlier... please add a similar unit test that ensures job service is getting added as an app property

controllers/profiles/common/app_properties_test.go
```diff
@@ -128,7 +128,7 @@ func Test_appPropertyHandler_WithServicesWithUserOverrides(t *testing.T) {
 	generatedProps, propsErr = properties.LoadString(props)
 	assert.NoError(t, propsErr)
 	assert.Equal(t, 9, len(generatedProps.Keys()))
-	assert.Equal(t, "http://"+platform.Name+"-"+DataIndexName+"."+platform.Namespace+"/processes", generatedProps.GetString(dataIndexServiceUrlProperty, ""))
+	assert.Equal(t, "http://"+platform.Name+"-"+DataIndexServiceName+"."+platform.Namespace+"/processes", generatedProps.GetString(dataIndexServiceUrlProperty, ""))
```
tchughesiv, line 131: missed this earlier... please add a similar unit test that ensures job service is getting added as an app property

controllers/profiles/common/app_properties.go
```diff
@@ -156,7 +194,7 @@ func (a *appPropertyHandler) withDataIndexServiceUrl() AppPropertyHandler {
 	if profiles.IsProdProfile(a.workflow) && dataIndexEnabled(a.platform) {
 		a.addDefaultMutableProperty(
 			dataIndexServiceUrlProperty,
-			fmt.Sprintf("%s://%s.%s/processes", dataIndexServiceUrlProtocol, GetDataIndexName(a.platform), a.platform.Namespace),
+			fmt.Sprintf("%s://%s.%s/processes", dataIndexServiceUrlProtocol, GetServiceName(a.platform.Name, DataIndexService), a.platform.Namespace),
```
tchughesiv, line 197: does a job service app property also need to be added here so that workflows know how to communicate to it?

tchughesiv (changes requested):

controllers/profiles/common/app_properties_test.go
tchughesiv: may want tests that check each configured service separately and then together

controllers/profiles/common/app_properties.go
jordigilh: Looks like it [has](https://kiegroup.github.io/kogito-docs/serverlessworkflow/latest/job-services/quarkus-extensions.html#kogito-addons-quarkus-jobs-knative-eventing) at least 3 from what I can gather by the data index property. 
```
mp.messaging.outgoing.kogito-job-service-job-request-events.connector=quarkus-http
mp.messaging.outgoing.kogito-job-service-job-request-events.url=${K_SINK:http://localhost:8280/v2/jobs/events}
mp.messaging.outgoing.kogito-job-service-job-request-events.method=POST
```
But, as it [happens](https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/162f3693844de04e94b447b29253eb3ed38057ad/controllers/profiles/common/app_properties.go#L44) with data index, only the `URL` property is added in the code for data index, the following being the equivalent for job service:
```
mp.messaging.outgoing.kogito-job-service-job-request-events.url
```

I will follow example from data index and add this last property if the job service is enabled

controllers/profiles/common/app_properties_test.go
jordigilh: Done in the same test

api/v1alpha08/sonataflowplatform_services_types.go
```diff
@@ -19,6 +19,9 @@ type ServicesPlatformSpec struct {
 	// Deploys the Data Index service for use by "prod" profile workflows.
 	// +optional
 	DataIndex *ServiceSpec `json:"dataIndex,omitempty"`
+	// Deploys the Job service for use by "prod" profile workflows.
+	// +optional
+	JobService *ServiceSpec `json:"job,omitempty"`
```
tchughesiv, line 24: didn't we talk about naming this field `jobService` ?
```go
	JobService *ServiceSpec `json:"jobService,omitempty"`
```

tchughesiv (changes requested):

api/v1alpha08/sonataflowplatform_services_types.go
jordigilh: good catch! I'll rename it as agreed to `jobService`

wmedvede (commented):
I've added some comments guys at the time we continue working in parallel to make the service work by using the PR sources.

good job in general!

controllers/platform/services.go
```diff
@@ -58,30 +60,72 @@ func (action *serviceAction) Handle(ctx context.Context, platform *operatorapi.S
 		return nil, err
 	}
 
-	if err := createDataIndexComponents(ctx, action.client, platform); err != nil {
-		return nil, err
+	if platform.Spec.Services.DataIndex != nil {
```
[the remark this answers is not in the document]
tchughesiv: agreed we should enhance the SonataFlowPlatform's status to account for service deployments.

controllers/platform/services.go
```diff
@@ -99,27 +98,9 @@ func createDataIndexDeployment(ctx context.Context, client client.Client, platfo
...
-		},
-		Resources: corev1.ResourceRequirements{
-			Limits: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("100m"),
-				corev1.ResourceMemory: resource.MustParse("256Mi"),
-			},
-		},
+		Image:          svc.GetServiceImageName(common.PersistenceTypeEphemeral),
+		Env:            svc.GetEnvironmentVariables(),
+		Resources:      svc.GetPodResourceRequirements(),
```
ricardozanini, line 103: `svc` is an alias for Kubernetes Services, can we change this since this context is different, please?

controllers/profiles/common/app_properties.go
```diff
@@ -38,34 +43,313 @@ import (
...
 )
 
-var immutableApplicationProperties = "quarkus.http.port=" + DefaultHTTPWorkflowPortIntStr.String() + "\n" +
-	"quarkus.http.host=0.0.0.0\n" +
-	// We disable the Knative health checks to not block the dev pod to run if Knative objects are not available
-	// See: https://kiegroup.github.io/kogito-docs/serverlessworkflow/latest/eventing/consume-produce-events-with-knative-eventing.html#ref-knative-eventing-add-on-source-configuration
-	"org.kie.kogito.addons.knative.eventing.health-enabled=false\n" +
-	"quarkus.devservices.enabled=false\n" +
-	"quarkus.kogito.devservices.enabled=false\n"
+type PlatformService interface {
```
ricardozanini, line 85: We are mixing Application properties with object creations in this file. This `PlatformService` must be within the package `platform` and not be exported.

controllers/profiles/common/app_properties.go
```diff
@@ -38,34 +43,313 @@ import (
...
+	DataIndexName                            = "data-index"
+	JobServiceName                           = "jobs-service"
+	PersistenceTypeEphemeral                 = "ephemeral"
+	PersistenceTypePostgressql               = "postgresql"
 	microprofileServiceCatalogPropertyPrefix = "org.kie.kogito.addons.discovery."
 	discoveryLikePropertyPattern             = "^\\${(kubernetes|knative|openshift):(.*)}$"
+
+	defaultDatabaseName   string = "sonataflow"
+	defaultPostgreSQLPort int    = 5432
+)
```
ricardozanini, line 70: There are too many properties here that belong to services. Can we please have an app properties handler for workflows and other for the services?

`platform` -> handler for Services
`profiles` -> handler for workflows

We might have a common module that defines a strategy for handling properties, but each module should have its own responsibility accordingly.

ricardozanini (changes requested):
Thank you for the refactoring, but can you kindly consider my comments? Many thanks!

controllers/platform/services.go
```diff
@@ -58,30 +55,32 @@ func (action *serviceAction) Handle(ctx context.Context, platform *operatorapi.S
 		return nil, err
 	}
 
-	if err := createDataIndexComponents(ctx, action.client, platform); err != nil {
-		return nil, err
+	if platform.Spec.Services.DataIndex != nil {
```
masayag, line 58: Should we get into this block if `platform.Spec.Services.DataIndex.Enabled == false`?

controllers/platform/services.go
```diff
@@ -58,30 +55,32 @@ func (action *serviceAction) Handle(ctx context.Context, platform *operatorapi.S
...
+		if err := createServiceComponents(ctx, action.client, platform, common.NewDataIndexService(platform)); err != nil {
+			return nil, err
+		}
+	}
+
+	if platform.Spec.Services.JobService != nil {
```
masayag, line 64: same here, Should we get into this block if `platform.Spec.Services.JobService.Enabled == false`?

controllers/platform/services.go
masayag: per golang convention, this can be simply `ps`  (shorten for PlatformService)

controllers/platform/services.go
```diff
@@ -99,27 +98,9 @@ func createDataIndexDeployment(ctx context.Context, client client.Client, platfo
 	liveProbe := readyProbe.DeepCopy()
 	liveProbe.ProbeHandler.HTTPGet.Path = common.QuarkusHealthPathLive
 	dataDeployContainer := &corev1.Container{
```
masayag, line 100: @ricardozanini is the `data` prefix related to the data-index from the origin function purpose or serves a different meaning here? If it ties to `data-index`, can we omit it from the name of this variable to make this function more generic?

controllers/platform/services.go
tchughesiv: it's correct as it is. the `Enabled` field does not control whether a service is deployed. instead, it controls whether its used by the "prod" workflows -
https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/a96add5c995496ef2d02f313114efed7d74a4725/api/v1alpha08/sonataflowplatform_services_types.go#L27-L29

controllers/platform/services.go
ricardozanini: Are you talking about the env var? I'm sure if I followed.

tchughesiv: i think he means, why is `data` part of the `dataDeployContainer` var. with @jordigilh's other changes, this could maybe be changed to `deployContainer`?

ricardozanini: +1

controllers/platform/services.go
masayag: Thank you for the reference.
In which case should this be set to false, assuming the workflow's profile is set to `prod`?

controllers/platform/services.go
masayag: Thank you @tchughesiv :-)

controllers/platform/services.go
tchughesiv: no, this flag only affects "prod" workflows. if it's a dev workflow, this flag won't even be checked by the operator. when a user elects to set `platform.spec.services.dataIndex.enabled` to `true` (which is the default), they are saying they want "prod" workflows to use this data index.

ricardozanini: I agree, once we have configs changed, we then rollout every workflow removing this property so then we scale it to 0. For this initial version, is fine the way it is.

tchughesiv: @ricardozanini it's worth asking the question though, to @masayag's point... is this a use-case we should be supporting anyway? What IS the use-case? When would an admin or dev need/want to "disable" these services? FWIW, the user could just as easily set the service replicas to 0 themselves. Is this "enabled" field necessary?

I'm good either way... just asking the Q.

controllers/profiles/common/app_properties.go
jordigilh: Done. I'm not sure if I got it right though. There are a lot of inter dependencies in the handler and the platform services that caused a import cycle problem when I tried to keep things together. Eventually I found a way to make a clear cut by defining a package inside `platform` and `profile` that allowed me to call the platform service properties inside the profile handler like it's being done now.

If what you expect is to duplicate the handler struct and functions for `platform` and `profiles`, with the caveat that only in `platform` it will inject the platform services properties, let me know.

controllers/profiles/common/app_properties.go
jordigilh: Moved inside the`platform` package.

controllers/platform/services.go
jordigilh: Done.

controllers/platform/services.go
ricardozanini: I keep asking these to myself. Let's reevaluate the need for it. I remember we discussed how to have the platform deploy the services and that's what we came up with. If we enable deployment, we should also disable it. We must have a response to irt or change the structure in a way that "disable" won't make sense.

controllers/platform/services/properties.go
```diff
@@ -0,0 +1,97 @@
...
+	databaseName := dbName
+	if len(postgresSpec.ServiceRef.DatabaseName) > 0 {
+		databaseName = postgresSpec.ServiceRef.DatabaseName
+	}
+	return fmt.Sprintf("%s://%s:%d/%s?search_path=%s", constants.PersistenceTypePostgreSQL, postgresSpec.ServiceRef.Name+"."+databaseNamespace, dataSourcePort, databaseName, databaseSchema)
+}
+
+// withDataIndexServiceUrl adds the property dataIndexServiceUrlProperty to the application properties.
+// See Service Discovery https://kubernetes.io/docs/concepts/services-networking/service/#dns
+func GenerateDataIndexApplicationProperties(workflow *operatorapi.SonataFlow, platform *operatorapi.SonataFlowPlatform) *properties.Properties {
```
ricardozanini, line 53: I think this comment is displaced now.

controllers/platform/services/properties.go
```diff
@@ -0,0 +1,97 @@
...
+		props.Set(
+			constants.DataIndexServiceURLProperty,
+			fmt.Sprintf("%s://%s.%s/processes", constants.DataIndexServiceURLProtocol, di.GetServiceName(), platform.Namespace),
+		)
+	}
+	return props
+}
+
+// withJobServiceURL adds the property 'mp.messaging.outgoing.kogito-job-service-job-request-events.url' to the application property
+func GenerateJobServiceApplicationProperties(workflow *operatorapi.SonataFlow, platform *operatorapi.SonataFlowPlatform) *properties.Properties {
```
ricardozanini, line 66: Same here regarding the comments.

ricardozanini (commented):
@wmedvede can you please take a second look too?

controllers/platform/services/properties.go
jordigilh: Done

jordigilh:
> @jordigilh does your PR descrip still match the changes you're making here? if not, could you update it? thanks!

Update.

controllers/platform/services.go
masayag: Maybe this was discussed before, so dropping some ideas here:

Perhaps the desired behavior here is more aligned with [k8s rolloutStrategy](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#deploymentstrategy-v1-apps). 
I'd explore the idea of replacing the `Enabled` flag with a `rolloutStrategy` property.
It leaves more room for introducing behavior types on CR change. In case of a change should be reflected in the workflows, the `rolloutStrategy` will determine the behavior. Few values to `rolloutStrategy`:
* Disabled - do nothing when platform CR is changed. this requires some out-of-sync between platform configuration to what is currently served by the workflows.
* Recreate - restart all with new configuration
* RollingUpdate - rolling the changes.
* Detached - detaching the service from workflow definition.

I'm also okay with not having it at all, but making the sync-status with the platform configuration visible is useful for debugging.
For once, I'd go with a single flag or strategy for all of the services and not per service.
Also, perhaps the future will knative service will not need this type of handling at all.

ricardozanini: @masayag that sounds reasonable, mind opening an issue here describing this behavior so we can discuss it more in-depth? We might have to implement it preferably before the cluster scope one. @tchughesiv wdyt about his proposal?

tchughesiv: I also think this all sounds reasonable, with `RollingUpdate` as the default when this new "flag" is not set. @ricardozanini I'm not sure the cluster-scoped changes would need to wait for this though... what what your thinking with that?

ricardozanini: Won't we replicate the spec in the cluster scope? I'd rather have the change once and propagate to the cluster-scoped requirements.

tchughesiv: no, we will not replicate the spec in cluster-scope. we previously discussed this and decided to only reference the chosen platform in spec and leave it at that. the referenced SonataFlowPlatform will be the single source of truth.

masayag: See https://github.com/apache/incubator-kie-kogito-serverless-operator/issues/321

controllers/platform/services/services.go
```diff
@@ -0,0 +1,298 @@
...
+					LocalObjectReference: secretRef,
+				},
+			},
+		},
+		{
+			Name:  "QUARKUS_DATASOURCE_DB_KIND",
+			Value: constants.PersistenceTypePostgreSQL,
+		},
+		{
+			Name:  "QUARKUS_HIBERNATE_ORM_DATABASE_GENERATION",
```
wmedvede, line 286: This env var don't apply for the jobs-service

wmedvede (commented):
LGTM, I've added some comments, thanks!

controllers/platform/services/properties.go
```diff
@@ -0,0 +1,98 @@
...
+		)
+	}
+	return props
+}
+
+// GenerateJobServiceApplicationProperties returns the application properties required for the Job Service to work in a production profile and job service's
+// service's spec field `Enabled` set to true
+func GenerateJobServiceApplicationProperties(workflow *operatorapi.SonataFlow, platform *operatorapi.SonataFlowPlatform) *properties.Properties {
+	props := properties.NewProperties()
+	if profiles.IsProdProfile(workflow) && jobServiceEnabled(platform) {
```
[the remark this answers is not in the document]
wmedvede: btw,  @jordigilh  @ricardozanini the comment above regarding the IsProdProfile check apply also for the data-index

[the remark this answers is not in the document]
ricardozanini: Yes, that should work independently.

controllers/platform/services/services.go
```diff
@@ -0,0 +1,298 @@
...
+	// GetServiceCmName returns the name of the configmap associated to the service
+	GetServiceCmName() string
+	// GetEnvironmentVariables returns the env variables to be injected to the service container
+	GetEnvironmentVariables() []corev1.EnvVar
+	// GetResourceLimits returns the pod's memory and CPU resource requirements
+	// Values for job service taken from
+	// https://github.com/parodos-dev/orchestrator-helm-chart/blob/52d09eda56fdbed3060782df29847c97f172600f/charts/orchestrator/values.yaml#L68-L72
+	GetPodResourceRequirements() corev1.ResourceRequirements
+	// GetReplicaCountForService Returns the default pod replica count for the given service
+	GetReplicaCount() int32
```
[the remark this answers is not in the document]
ricardozanini: @wmedvede well noted, we can't scale Jobs Service to more than one replica.

controllers/platform/services/properties.go
```diff
@@ -0,0 +1,98 @@
...
+	operatorapi "github.com/apache/incubator-kie-kogito-serverless-operator/api/v1alpha08"
+	"github.com/apache/incubator-kie-kogito-serverless-operator/controllers/profiles"
+	"github.com/apache/incubator-kie-kogito-serverless-operator/controllers/profiles/common/constants"
+
+	"github.com/magiconair/properties"
+)
+
+func generateReactiveURL(postgresSpec *operatorapi.PersistencePostgreSql, schema string, namespace string, dbName string, port int) string {
+	if len(postgresSpec.JdbcUrl) > 0 && strings.Contains(postgresSpec.JdbcUrl, "currentSchema=") {
+		return strings.Replace(strings.TrimPrefix(postgresSpec.JdbcUrl, "jdbc:"), "currentSchema=", "search_path=", -1)
```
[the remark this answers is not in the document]
jordigilh: so `currentSchema` is optional then? Shall I just replace it when found, otherwise append `[?]search_path=<schema>`?
```
postgresql://postgres:5432/sonataflow?search_path=<schema>
```

controllers/platform/services/services.go
jordigilh: ack

controllers/profiles/common/constants/platform_services.go
```diff
@@ -0,0 +1,51 @@
...
+	JobServiceStatusChangeEventsProperty    = "kogito.jobs-service.http.job-status-change-events"
+	JobServiceStatusChangeEventsURL         = "mp.messaging.outgoing.kogito-job-service-job-status-events-http.url"
+	JobServiceURLProtocol                   = "http"
+	JobServiceDataSourceReactiveURLProperty = "quarkus.datasource.reactive.url"
+
+	KogitoServiceUrlProperty    = "kogito.service.url"
+	KogitoServiceUrlProtocol    = "http"
+	KafkaSmallRyeHealthProperty = "quarkus.smallrye-health.check.\"io.quarkus.kafka.client.health.KafkaHealthCheck\".enabled"
+
+	MicroprofileServiceCatalogPropertyPrefix = "org.kie.kogito.addons.discovery."
```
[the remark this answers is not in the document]
jordigilh: ack

controllers/profiles/common/constants/platform_services.go
```diff
@@ -0,0 +1,51 @@
...
+	DiscoveryLikePropertyPattern             = "^\\${(kubernetes|knative|openshift):(.*)}$"
```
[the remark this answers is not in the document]
jordigilh: ack

controllers/profiles/common/properties/application.go
```diff
@@ -27,46 +27,30 @@ import (
...
+var (
+	immutableApplicationProperties = fmt.Sprintf("quarkus.http.port=%d\n"+
+		"quarkus.http.host=0.0.0.0\n"+
+		// We disable the Knative health checks to not block the dev pod to run if Knative objects are not available
+		// See: https://kiegroup.github.io/kogito-docs/serverlessworkflow/latest/eventing/consume-produce-events-with-knative-eventing.html#ref-knative-eventing-add-on-source-configuration
+		"org.kie.kogito.addons.knative.eventing.health-enabled=false\n"+
+		"quarkus.devservices.enabled=false\n"+
+		"quarkus.kogito.devservices.enabled=false\n", constants.DefaultHTTPWorkflowPortInt)
+
+	discoveryLikePropertyExpr                    = regexp.MustCompile(constants.DiscoveryLikePropertyPattern)
```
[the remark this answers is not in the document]
jordigilh: ack

controllers/platform/services/services.go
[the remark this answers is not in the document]
jordigilh: ok so it should always set to 1 regardless of what it's elswhere defined, right?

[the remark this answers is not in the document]
jordigilh: Done, hardcoded to return always 1

controllers/profiles/common/properties/application.go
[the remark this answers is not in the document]
jordigilh: Done, moved constant to this file and made it private

controllers/profiles/common/constants/platform_services.go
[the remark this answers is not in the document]
jordigilh: Moved constant to the `workflows.go` file inside the constants package

controllers/profiles/common/constants/platform_services.go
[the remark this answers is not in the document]
jordigilh: Moved constant to the `workflows.go` file inside the constants package

controllers/platform/services/services.go
jordigilh: Done.

controllers/platform/services/properties.go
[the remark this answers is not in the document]
jordigilh: Removed references to the workflow in both data index and job service application properties generation.

controllers/platform/services/properties.go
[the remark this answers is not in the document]
jordigilh: Updated code to parse the URL and generate it based on the contents of the parsed object.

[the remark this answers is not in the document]
wmedvede: yes, current schema is optional, and if no currentSchema, no search_path

controllers/platform/services/properties.go
```diff
@@ -0,0 +1,131 @@
...
+				}
+			}
+		} else if v, ok := kv["currentSchema"]; ok {
+			for _, val := range v {
+				if len(val) != 0 {
+					spv = v[0]
+				}
+			}
+		}
+		return fmt.Sprintf("%ssearch_path=%s", ret, spv), nil
```
wmedvede, line 61: when we reach this point, if we couldn't infer the schema from the JdbcUrl, we don't set it.

wmedvede (commented):
One last nitpick that I think makes sense, and then I'll retest all this with the different variants  in https://github.com/wmedvede/techpreview2/tree/test-platform-services

controllers/platform/services/properties.go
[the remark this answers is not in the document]
tchughesiv: @jordigilh can you explain what was changed here and why?

controllers/profiles/common/properties/application.go
```diff
@@ -143,31 +131,18 @@ func (a *appPropertyHandler) Build() string {
...
-		kogitoServiceUrl = fmt.Sprintf("%s://%s", kogitoServiceUrlProtocol, a.workflow.Name)
+		kogitoServiceUrl = fmt.Sprintf("%s://%s", constants.KogitoServiceUrlProtocol, a.workflow.Name)
 	}
-	return a.addDefaultMutableProperty(kogitoServiceUrlProperty, kogitoServiceUrl)
-}
-
-// withDataIndexServiceUrl adds the property dataIndexServiceUrlProperty to the application properties.
-// See Service Discovery https://kubernetes.io/docs/concepts/services-networking/service/#dns
-func (a *appPropertyHandler) withDataIndexServiceUrl() AppPropertyHandler {
-	if profiles.IsProdProfile(a.workflow) && dataIndexEnabled(a.platform) {
```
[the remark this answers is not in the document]
tchughesiv: if we're now configuring ALL workflows with these services, then many of api descriptions and references to this functionality would need to be changed as well.

controllers/platform/services/properties.go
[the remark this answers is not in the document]
wmedvede: @tchughesiv This change is because in order for the workflows to work, (when configured) we need the jobs-service and data-index to be deployed first, but also, it might occur, that we push the platform into the namespace, and then, we push the workflows. So, what we are looking for with this change is that the deployment of the services are not dependent on any particular workflow.
That's basically the aim of this change.

controllers/profiles/common/properties/application.go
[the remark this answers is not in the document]
ricardozanini: Only workflows not in `dev` profile in the given `SonataFlowPlatform` should be configured to communicate with Jobs Service and Data Index.

This is the use case.

controllers/platform/services/properties.go
jordigilh: If it can't infer the schema, I'll return the URL without the `search_path` parameter. Is that correct?

controllers/platform/services/properties.go
```diff
@@ -0,0 +1,131 @@
...
+			if ok {
+				ret = fmt.Sprintf("%s%s:%s@", ret, u.User.Username(), p)
+			}
+		}
+		ret = fmt.Sprintf("%s%s%s?", ret, u.Host, u.Path)
+		kv, err := url.ParseQuery(u.RawQuery)
+		if err != nil {
+			return "", err
+		}
+		spv := schema
```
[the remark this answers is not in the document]
jordigilh: When the schema is not found in the URL, is that OK to return the JDBC URL without any parameter?
Example:
Source URL `jdbc:postgres://host.com:5432/path?k=v#f` -> Reactive URL ->`postgres://host.com:5432/path`

controllers/profiles/common/properties/application.go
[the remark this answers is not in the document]
jordigilh: As agreed, I'll revert the change and add back the conditional to check if the workflow is not nil and if the profile is not DEV.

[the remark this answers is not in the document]
jordigilh: Done to check if the workflow does not contain the `Dev` profile instead of checking if it has the `Prod` profile

controllers/platform/services/properties.go
jordigilh: Done. PTAL

jordigilh:
Let me know if you want me to squash the commits before merging.

controllers/platform/services/services.go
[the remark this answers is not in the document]
jordigilh: Resolving this comment as it has already answered earlier

wmedvede:
Hello Guys @jordigilh @tchughesiv @ricardozanini , I'll add some test results here. (the last commits added by @jordigilh were used)

**Note:**
In general I have added reproducers for the tests cases here: https://github.com/wmedvede/techpreview2/tree/test-platform-services, feel to clone and give it a try locally

**Case 1:** 
We deploy the platform with jobs service + postgresql persistence (no workflows):

**reproducer here:** https://github.com/wmedvede/techpreview2/blob/1a2cc37f4f9a9c1423db873da847ee409a576e4d/platforms/jobservice_as_platform_service_postgresql/Readme.txt

What we can see is that the jobs service tries to start, but, no reactive url is created, and thus it fails.
These are the options in my opinion:

**Option 1)** we don't let it start, since we have no workflows yet.

**Option 2)** we let it start, in that case, we need the reactive datasource URL
I think we have discussed already in favor of option 2.

Screenshot from 2023-12-14 10-19-03

wmedvede:
**Case2:**
Given the platform of Case1, in a later point in time I try to deploy a workflow.
Reproducer here:

These are the generated properties for the workflow:

Screenshot from 2023-12-14 11-16-51

**here we can see:**

quarkus.datasource.reactive.url = postgresql://postgres:5432/sonataflow?search_path=jobs-service  (is not for the workflow, but for the jobs-service)

quarkus.smallrye-health.check.\"org.kie.kogito.jobs.service.messaging.http.health.knative.KSinkInjectionHealthCheck\".enabled = false   (is not for the workflow, but for the jobs-service)

**Some missing properties for the workflow: (These ones are, this was my fault, I forgot to mention this ones :(  )**

mp.messaging.outgoing.kogito-job-service-job-request-events.connector=quarkus-http

kogito.events.processdefinitions.enabled=false
kogito.events.usertasks.enabled=false
kogito.events.variables.enabled=false

**Remainder:** 
If we detect that the Data Index is present, then we must also generate this properties for the workflow. I haven't tested this last,

mp.messaging.outgoing.kogito-processinstances-events.url=http://<data index service correct url>/processes

controllers/profiles/common/properties/application_test.go
```diff
@@ -238,3 +182,140 @@ func doTestGenerateMicroprofileServiceCatalogProperty(t *testing.T, serviceUri s
...
+	assert.Equal(t, 8, len(generatedProps.Keys()))
+	assert.Equal(t, "", generatedProps.GetString(constants.DataIndexServiceURLProperty, ""))
+	assert.Equal(t, "", generatedProps.GetString(constants.JobServiceURLProperty, ""))
+	assert.Equal(t, "", generatedProps.GetString(constants.JobServiceKafkaSinkInjectionHealthCheck, ""))
+	assert.Equal(t, "", generatedProps.GetString(constants.JobServiceDataSourceReactiveURLProperty, ""))
+	assert.Equal(t, "", generatedProps.GetString(constants.JobServiceStatusChangeEventsProperty, ""))
+	assert.Equal(t, "", generatedProps.GetString(constants.JobServiceStatusChangeEventsURL, ""))
+
+	// check that service app properties are being properly set
+	generatedProps, propsErr = properties.LoadString(NewServiceAppPropertyHandler(platform).WithUserProperties(userProperties).Build())
```
dmartinol, line 268: in the business code (`services.go`) there is no such case where `NewServiceAppPropertyHandler` is decorated with `WithUserProperties`. Is this a correct population of the test data?
According to what I undertstood from @wmedvede
> there are no DI or JS properties that requires user intervention

so I would expect that there's no way to have `property1`, `property2` in the `generatedProps` (and `kogito.service.url` either), and the len should be 6 instead of 9

tchughesiv: > there are no DI or JS properties that requires user intervention
while there may not be properties that "require" user intervention... in my opinion, the operator shouldn't be in the business of limiting the user wrt property configs either. we already allow the user the modify just about everything in the DI & JS PodSpec... application properties should be no different. this kind of functionality allows for greater adoption and success in future production-like envs.

jordigilh:
> Hello Guys @jordigilh @tchughesiv @ricardozanini , I'll add some test results here. (the last commits added by @jordigilh were used)
> 
> **Note:** In general I have added reproducers for the tests cases here: https://github.com/wmedvede/techpreview2/tree/test-platform-services, feel to clone and give it a try locally
> 
> **Case 1:** We deploy the platform with jobs service + postgresql persistence (no workflows):
> 
> **reproducer here:** https://github.com/wmedvede/techpreview2/blob/1a2cc37f4f9a9c1423db873da847ee409a576e4d/platforms/jobservice_as_platform_service_postgresql/Readme.txt
> 
> What we can see is that the jobs service tries to start, but, no reactive url is created, and thus it fails. These are the options in my opinion:
> 
> **Option 1)** we don't let it start, since we have no workflows yet.
> 
> **Option 2)** we let it start, in that case, we need the reactive datasource URL I think we have discussed already in favor of option 2.
> 
> Screenshot from 2023-12-14 10-19-03
> 
> These are the generated properties:
> 
> Screenshot from 2023-12-14 11-09-21

There are 2 issues I noticed in the logs:
* The controller keeps dumping this error due to a schema validation in the persistence struct for postgreSQL:
https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/3b5108ca8c352cb3d81fd736f3e116ed21e66a3f/api/v1alpha08/sonataflowplatform_services_types.go#L47C4-L48
```
E1214 16:50:41.630051       1 defaults.go:79] "Error updating the BuildPlatform" err="SonataFlowPlatform.sonataflow.org \"sonataflow-platform\" is invalid: spec.services.jobService.persistence.postgresql: Too many: 3: must have at most 2 items"
```
It's fixed by increasing the `maxProperty` value to 3. @tchughesiv do you know why it was limited to 2 max in the first place? Any reason for the validation? Can it be increased to 3? 

* The job service app properties are not added because the code that adds these properties is never called. The code is executed only when the a sonataflow CR is created, but not when the service is created. In this case should it add the same properties to the configmap as it does for a sonataflow CR? Or should it be a subset of those (reactiveURL and something else?). @wmedvede  please advise.

jordigilh:
@wmedvede these are the properties generated for the job service once I reuse the same code for the app properties handler:
```
mp.messaging.outgoing.kogito-processinstances-events.url = http://sonataflow-platform-data-index-service.default/processes
mp.messaging.outgoing.kogito-job-service-job-request-events.url = http://sonataflow-platform-jobs-service.default/v2/jobs/events
quarkus.smallrye-health.check."org.kie.kogito.jobs.service.messaging.http.health.knative.KSinkInjectionHealthCheck".enabled = false
quarkus.datasource.reactive.url = postgresql://postgres:5432/sonataflow?search_path=jobs-service
quarkus.smallrye-health.check."io.quarkus.kafka.client.health.KafkaHealthCheck".enabled = false
quarkus.http.port = 8080
quarkus.http.host = 0.0.0.0
org.kie.kogito.addons.knative.eventing.health-enabled = false
quarkus.devservices.enabled = false
quarkus.kogito.devservices.enabled = false
mp.messaging.outgoing.kogito-job-service-job-request-events.connector = quarkus-http
kogito.events.processdefinitions.enabled = false
kogito.events.usertasks.enabled = false
kogito.events.variables.enabled = false
```

Can you confirm that all keys for the job service are accounted for?

tchughesiv:
> It's fixed by increasing the `maxProperty` value to 3. @tchughesiv do you know why it was limited to 2 max in the first place? Any reason for the validation? Can it be increased to 3?

No, it should not be increased to 3. The validation is set this way because `ServiceRef` and `JdbcUrl` should not be set simultaneously. It's an either/or scenario. `SecretRef` is a required field.

jordigilh:
> > It's fixed by increasing the `maxProperty` value to 3. @tchughesiv do you know why it was limited to 2 max in the first place? Any reason for the validation? Can it be increased to 3?
> 
> No, it should not be increased to 3. The validation is set this way because `ServiceRef` and `JdbcUrl` should not be set simultaneously. It's an either/or scenario. `SecretRef` is a required field.
> 
> @jordigilh can you show me the platform CR you're trying to use?

The problem is that the `ServiceRef` field is not a pointer and it's a structure, so it is populated with default values and that's good enough for the schema validation to be considered as populated. That's why, when providing the JDBC URL, it considers all 3 fields as populated and fails the validation.

By changing the `ServiceRef` to be a pointer, and adjusting the code to take that into account, solves the issue without having to change the limits:

```
	// Service reference to postgresql datasource. Mutually exclusive to jdbcUrl.
	// +optional
	ServiceRef *PostgreSqlServiceOptions `json:"serviceRef,omitempty"`
```

You can reproduce the error by creating this test case:
https://github.com/wmedvede/techpreview2/tree/test-platform-services/platforms/jobservice_as_platform_service_postgresql

tchughesiv:
> By changing the `ServiceRef` to be a pointer, and adjusting the code to take that into account, solves the issue without having to change the limits:
@jordigilh that makes sense... i think i only tested this in the OpenShift UI so this didn't bite me.

ricardozanini:
> The job service app properties are not added because the code that adds these properties is never called. The code is executed only when the a sonataflow CR is created, but not when the service is created. In this case should it add the same properties to the configmap as it does for a sonataflow CR? Or should it be a subset of those (reactiveURL and something else?).

1. When deploying Jobs Service or Data Index, the `SonataFlowPlatform` controller must set the config properties for each service accordingly for them to work independently.
2. During the `SonataFlow` controller reconciliation, if there's a `SonataFlowPlatform` with Jobs Service and DI available in the current context, it should have the workflow application properties being updated to add the required properties for the workflow to connect and send events to those services.

That's the use case.

jordigilh:
>     1. When deploying Jobs Service or Data Index, the `SonataFlowPlatform` controller must set the config properties for each service accordingly for them to work independently.
>     2. During the `SonataFlow` controller reconciliation, if the workflow is not `dev` and if there's a `SonataFlowPlatform` with Jobs Service and DI available in the current context, it should have the workflow application properties being updated to add the required properties for the workflow to connect and send events to those services.
> 
> 
> That's the use case.

Should the properties in each service be any different for the service itself or for the deployment triggered by a sonataflow CR instance? In other words: the properties specific for creating the job service are all included in the properties for a sonataflow CR? or a subset of those?

Right now I consider them to be the same.

tchughesiv:
> > ```
> > 1. When deploying Jobs Service or Data Index, the `SonataFlowPlatform` controller must set the config properties for each service accordingly for them to work independently.
> > 2. During the `SonataFlow` controller reconciliation, if the workflow is not `dev` and if there's a `SonataFlowPlatform` with Jobs Service and DI available in the current context, it should have the workflow application properties being updated to add the required properties for the workflow to connect and send events to those services.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That's the use case.
> 
> Should the properties in each service be any different for the service itself or for the deployment triggered by a sonataflow CR instance? In other words: the properties specific for creating the job service are all included in the properties for a sonataflow CR? or a subset of those?
> 
> Right now I consider them to be the same.

@jordigilh service properties will be different than workflow properties. sure, they may share a few similar configs but there will be configs that only services use... and configs that only the workflows will use. you should be able to reference what was already done w/ DI vs the workflow properties.. and then modify to fit the needs of job service. e.g. the `NewServiceAppPropertyHandler`.

ricardozanini:
Hi @jordigilh 

> in other words: the properties specific for creating the job service are all included in the properties for a sonataflow CR? or a subset of those?

> Right now I consider them to be the same.

No, they are not the same. I think it was there that things messed a little bit. Take a look at this repo we shared at the beginning of the development:

https://github.com/flows-examples/techpreview2/tree/main/infra/dataindex_and_jobservice

You can see the JS Deployment and the `application.properties` for it there. That's exactly what we were expecting for this PR to do.

And then, if you see the use case:
https://github.com/flows-examples/techpreview2/tree/main?tab=readme-ov-file#use-case-3

It's explained there how it is supposed to be working. The workflow has yet another set of properties:
https://github.com/flows-examples/techpreview2/blob/main/workflows/sonataflow-callbackstatetimeouts-persistence/02-configmap_callbackstatetimeouts-props

That's why I asked to separate the props in the first place last week between the services and the workflow. I thought it was clear since these use cases were already working and defined in that repo. I'm sorry if I made an incorrect assumption.

jordigilh:
> Hi @jordigilh
> 
> > in other words: the properties specific for creating the job service are all included in the properties for a sonataflow CR? or a subset of those?
> 
> > Right now I consider them to be the same.
> 
> No, they are not the same. I think it was there that things messed a little bit. Take a look at this repo we shared at the beginning of the development:
> 
> https://github.com/flows-examples/techpreview2/tree/main/infra/dataindex_and_jobservice
> 
> You can see the JS Deployment and the `application.properties` for it there. That's exactly what we were expecting for this PR to do.
> 
> And then, if you see the use case: https://github.com/flows-examples/techpreview2/tree/main?tab=readme-ov-file#use-case-3
> 
> It's explained there how it is supposed to be working. The workflow has yet another set of properties: https://github.com/flows-examples/techpreview2/blob/main/workflows/sonataflow-callbackstatetimeouts-persistence/02-configmap_callbackstatetimeouts-props
> 
> That's why I asked to separate the props in the first place last week between the services and the workflow. I thought it was clear since these use cases were already working and defined in that repo. I'm sorry if I made an incorrect assumption.

Ok, so for the sake of clarity because I'm not yet sure what should be in where:
* When deploying the job service alone, the properties that should go specific for the job service are as follows (just adding the key here for simplicity):
  * `kogito.service.url`
  * `quarkus.datasource.jdbc.url`
  * `quarkus.datasource.reactive.url`
  * `kogito.jobs-service.http.job-status-change-events`
  * `mp.messaging.outgoing.kogito-job-service-job-status-events-http.url`
  * `quarkus.smallrye-health.check."org.kie.kogito.jobs.service.messaging.http.health.knative.KSinkInjectionHealthCheck".enabled`
  * When Data Index is deployed with a nonDev profile and with the enabled field set to `true`, additional properties to be added:
    * `kogito.jobs-service.http.job-status-change-events`
    * `mp.messaging.outgoing.kogito-job-service-job-status-events-http.url`
* When deploying a workflow, these are the keys expected generated by the job service with a postgreSQL DB for persistence:
  * `mp.messaging.outgoing.kogito-job-service-job-request-events.connector`
  * `mp.messaging.outgoing.kogito-job-service-job-request-events.url`

I can't find any property shared between the job service deployment and a workflow deployment.
From the [application.properties](https://github.com/flows-examples/techpreview2/blob/main/workflows/sonataflow-callbackstatetimeouts-persistence/02-configmap_callbackstatetimeouts-props.yaml) in the example @ricardozanini provided, I can see only 3 of the 4 properties that @wmedvede mentioned earlier in the case of deployment of a workflow: one listed in the job service section and the other 2 outside of it, which would mean they're out of scope of this PR:` kogito.events.usertasks.enabled`, `kogito.events.variables.enabled`. Missing is `kogito.events.processdefinitions.enabled`.

Is this configuration correct?

wmedvede:
Hi @jordigilh , I'll try to answer to your question above regarding the properties here:

**Jobs Service specific properties:**

kogito.service.url (cluster dns named url to self e.g. http://jobs-service-postgresql)
quarkus.datasource.jdbc.url
quarkus.datasource.reactive.url
health.check."org.kie.kogito.jobs.service.messaging.http.health.knative.KSinkInjectionHealthCheck".enabled=false

**Then, if data-index is present in the platform, we also add**

kogito.jobs-service.http.job-status-change-events=true
#cluster dns of the data index, which operator should know. example below
mp.messaging.outgoing.kogito-job-service-job-status-events-http.url=http://data-index-service-postgresql/jobs

jordigilh:
```
kogito.service.url (cluster dns named url to self e.g. http://jobs-service-postgresql/)
quarkus.datasource.jdbc.url
quarkus.datasource.reactive.url
health.check."org.kie.kogito.jobs.service.messaging.http.health.knative.KSinkInjectionHealthCheck".enabled=false
```
@wmedvede I noticed that the value for this property is already added as part of the environment values when defining the persistence. Should it be duplicated in the `application.properties`?

So far, only 2 properties seem to be generated from the job service itself, and that's the reactive URL and the health check. The kogito service URL is already added and the JDBC URL is added in the environment.

wmedvede:
@jordigilh 

**Properties for the workflows:**

kogito.service.url (we always set, it's the cluster dns url to my self, the operator was already doing it for the workflows)

**This two properties are needed to connect with the jobs service and we **must always set**** (no matter if jobs service is present or not)
mp.messaging.outgoing.kogito-job-service-job-request-events.connector=quarkus-http
mp.messaging.outgoing.kogito-job-service-job-request-events.url= http://jobs-service-postgresql/v2/jobs/events

If the jobs service is **NOT installed**, we still set the url, and we invent it :), to http://jobs-service/v2/jobs/events, np, we can manage in another PR to fine tune it. And, if the jobs service is installed we set the correct cluster dns url)

**Properties to connect with the data-index:** 

**If data index is not present we set this:** (all in false)
kogito.events.processinstances.enabled=false
kogito.events.processdefinitions.enabled=false
kogito.events.usertasks.enabled=false
kogito.events.variables.enabled=false


**If data index is present then we set this instead:**
kogito.events.processinstances.enabled=true  **(only this in true)**
kogito.events.processdefinitions.enabled=false
kogito.events.usertasks.enabled=false
kogito.events.variables.enabled=false

**And, since data index is present we add this one with the correct data  index cluster dns**

mp.messaging.outgoing.kogito-processinstances-events.url=http://data-index-service-postgresql/processes

wmedvede:
> quarkus.datasource.jdbc.url
@jordigilh  Your observation is correct, the quarkus.datasource.jdbc.url we are already setting in the env vars, **so we don't need to** set as property, It was added there for the example. But yeah, your comment is correct, well noted! , see that things like the database user and password we also set as env vars.

tchughesiv:
> If the jobs service is **NOT installed**, we still set the url, and we invent it :), to http://jobs-service/v2/jobs/events

This is dangerous as a service named `jobs-service` could exist within the namespace. are you sure this must be set? is there another, safer way to do this?

ricardozanini:
> > If the jobs service is **NOT installed**, we still set the url, and we invent it :), to http://jobs-service/v2/jobs/events
> 
> @wmedvede This is dangerous as a service named `jobs-service` could exist within the namespace. are you sure this must be set? is there a different, safer way to do this? maybe `http://localhost/v2/jobs/events` ?

The service URL is based on the objects we created in the namespace. Given that the default naming of the Deployment and Service are `jobs-service`, the DNS will resolve to `jobs-service.svc.cluster.local`, correct? 

So your concern is valid, but we will have a problem deploying the jobs-service before getting there since the deployment name will already be taken. So we need to fix the naming clash while creating these objects.

We can table this discussion for the moment and open a follow-up task to deal with the naming clash for our managed support services.

wmedvede:
Hello Guys, @jordigilh @ricardozanini @jordigilh here goes some test results with the latest commits:

**Jobs service only**

I think this case is working better now after the last commits. Jobs service is deployed and if we add workflows the properties are set, we still have some errors when we create a WF instance, but we can consider out scope of this PR.

See here please:  https://github.com/wmedvede/techpreview2/tree/test-platform-services/platforms/jobservice_as_platform_service_postgresql_openshift
 

**Data index only**

Don't work, see here please:

https://github.com/wmedvede/techpreview2/tree/test-platform-services/platforms/dataindex_as_platform_service_postgresql_openshift


**Jobs Service and Data Index **

Don't work, see here please: https://github.com/wmedvede/techpreview2/tree/test-platform-services/platforms/dataindex_and_jobservice_as_platform_service_postgresql_openshift


I think both cases well work better if we remove the wrong property 

mp.messaging.outgoing.kogito-processinstances-events.url =
    http://sonataflow-platform-data-index-service.test-platform3-oc/processes

tchughesiv:
@wmedvede @jordigilh `mp.messaging.outgoing.kogito-processinstances-events.url` should only be configured within the workflow properties when DI is enabled in the platform. with this PR it now appears it's being added to the DI/JS service properties as well. this is the result of the property refactoring that's been done.

`mp.messaging.outgoing.kogito-processinstances-events.url` && `mp.messaging.outgoing.kogito-job-service-job-request-events.url` are **client-side** configs and should only be set in the workflow properties when the corresponding service is enabled in the platform.

wmedvede:
> @wmedvede @jordigilh `mp.messaging.outgoing.kogito-processinstances-events.url` should only be configured within the workflow properties when DI is enabled in the platform. with this PR it now appears it's being added to the DI/JS service properties as well. this is the result of the property refactoring that's been done.
> 
> `mp.messaging.outgoing.kogito-processinstances-events.url` && `mp.messaging.outgoing.kogito-job-service-job-request-events.url` are **client-side** configs and should only be set in the workflow properties when the corresponding service is enabled in the platform.

yep, I think as soon this prop is not added any more we'll get both services starting.

ricardozanini:
@tchughesiv @wmedvede 

> If the jobs service is NOT installed, we still set the url, and we invent it :), to http://jobs-service/v2/jobs/events, np, we can manage in another PR to fine tune it. And, if the jobs service is installed we set the correct cluster dns url)

@tchughesiv I thought you meant the property on the workflow side to connect to JS. Yes, in this case, a property set in JS should be pointed to localhost instead within the same pod. `localhost` should do.

ricardozanini:
@jordigilh the e2e ` should successfully deploy the orderprocessing workflow in devmode and verify if it's running` is failing. You can run these from your local environment using `./local/hack/run-e2e.sh`.

jordigilh:
> ./local/hack/run-e2e.

Thanks! but I ended up running just `make test-e2e` and it worked 😄 .

Based on what I can gather from the pod's logs, it's failing due to an exception at startup:
```java
orderprocessing-5d586b99bb-vq7rp ERROR [io.quarkus.runtime.Application:-1] (Quarkus Main Thread) Failed to start application (with profile [dev]): java.lang.IllegalArgumentException: Unsupported scheme for the http connector in URL: services.v1/event-listener
	at io.quarkus.reactivemessaging.http.runtime.HttpSink.<init>(HttpSink.java:64)
	at io.quarkus.reactivemessaging.http.runtime.QuarkusHttpConnector.getSubscriberBuilder(QuarkusHttpConnector.java:119)
	at io.quarkus.reactivemessaging.http.runtime.QuarkusHttpConnector_Subclass.getSubscriberBuilder$$superforward1(Unknown Source)
	at io.quarkus.reactivemessaging.http.runtime.QuarkusHttpConnector_Subclass$$function$$3.apply(Unknown Source)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:54)
	at io.quarkus.smallrye.reactivemessaging.runtime.devmode.DevModeSupportConnectorFactoryInterceptor.intercept(DevModeSupportConnectorFactoryInterceptor.java:67)
	at io.quarkus.smallrye.reactivemessaging.runtime.devmode.DevModeSupportConnectorFactoryInterceptor_Bean.intercept(Unknown Source)
	at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:42)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:50)
	at io.quarkus.smallrye.reactivemessaging.runtime.DuplicatedContextConnectorFactoryInterceptor.intercept(DuplicatedContextConnectorFactoryInterceptor.java:37)
	at io.quarkus.smallrye.reactivemessaging.runtime.DuplicatedContextConnectorFactoryInterceptor_Bean.intercept(Unknown Source)
	at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:42)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.perform(AroundInvokeInvocationContext.java:41)
	at io.quarkus.arc.impl.InvocationContexts.performAroundInvoke(InvocationContexts.java:33)
	at io.quarkus.reactivemessaging.http.runtime.QuarkusHttpConnector_Subclass.getSubscriberBuilder(Unknown Source)
	at io.quarkus.reactivemessaging.http.runtime.QuarkusHttpConnector_ClientProxy.getSubscriberBuilder(Unknown Source)
	at io.smallrye.reactive.messaging.providers.impl.ConnectorFactories.lambda$wrap$5(ConnectorFactories.java:84)
	at io.smallrye.reactive.messaging.providers.impl.ConfiguredChannelFactory.createSubscriber(ConfiguredChannelFactory.java:191)
	at io.smallrye.reactive.messaging.providers.impl.ConfiguredChannelFactory.register(ConfiguredChannelFactory.java:146)
	at io.smallrye.reactive.messaging.providers.impl.ConfiguredChannelFactory.initialize(ConfiguredChannelFactory.java:107)
	at io.smallrye.reactive.messaging.providers.impl.ConfiguredChannelFactory_ClientProxy.initialize(Unknown Source)
```
I've checked the [configmap](https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/3b5108ca8c352cb3d81fd736f3e116ed21e66a3f/test/testdata/order-processing/04_v1_configmap_properties.yaml#L25) and there's a property added with that value:
```
mp.messaging.outgoing.kogito_outgoing_stream.url = ${kubernetes:services.v1/event-listener}
```

Here is the whole list of properties defined for the test:

```
kogito.events.processdefinitions.enabled = false
kogito.events.processinstances.enabled = false
kogito.events.usertasks.enabled = false
kogito.events.variables.enabled = false
kogito.service.url = http://orderprocessing.sonataflow-operator-system
mp.messaging.outgoing.kogito-job-service-job-request-events.connector = quarkus-http
mp.messaging.outgoing.kogito-job-service-job-request-events.url = http://localhost/v2/jobs/events
mp.messaging.outgoing.kogito_outgoing_stream.url = ${kubernetes:services.v1/event-listener}
org.kie.kogito.addons.knative.eventing.health-enabled = false
quarkus.devservices.enabled = false
quarkus.http.host = 0.0.0.0
quarkus.http.port = 8080
quarkus.kogito.devservices.enabled = false
quarkus.log.level = INFO
```

I'm unfamiliar with the cause of this exception. Does it ring any bells to you?

ricardozanini:
@wmedvede look at @jordigilh's comments. The image doesn't have the runtimes fabric8 discovery anymore, so the properties in this test case must be updated. 

@jordigilh I'll fix it in a separate PR, you can ignore it for the moment. After a meeting I'm having now, I'll fix it.

wmedvede:
> ```java
> orderprocessing-5d586b99bb-vq7rp ERROR [io.quarkus.runtime.Application:-1] (Quarkus Main Thread) Failed to start application (with profile [dev]): java.lang.IllegalArgumentException: Unsupported scheme for the http connector in URL: services.v1/event-listener
> ```

Hi @jordigilh , this PR https://github.com/apache/incubator-kie-kogito-serverless-operator/pull/332 should fix the issue regarding the E2E tests. No you'll find something like the screenshot blow in the WF properties, and the event-listener address is resolved well.

Screenshot from 2023-12-19 18-03-51

ricardozanini:
e2e fixed here: https://github.com/apache/incubator-kie-kogito-serverless-operator/pull/332

wmedvede (approved):
After the final local tests on openshift I can see the following scenarios working

**1) jobs-service and data-index deployed with postgresql + a later workflow deployment**
Services are started well and gets positive health checks, and has the expected configurations. Similar results for the workflow.

**2) jobs-service postgresql + a later workflow deployment**
Job service is started well and gets positive health checks, and has the expected configurations. Similar results for the workflow.

**3) data-index deployed with postgresql + a later workflow deployment**
Data index service is started well getting positive health checks and, and has expected configurations, similar for the workflows.

I conclude that we can merge, other fine tunnings or missing tests can be done in a follow-up PR.

minkube deployment checks where was done by @jordigilh 

Good job!

@ricardozanini let's merge!

tchughesiv (approved):
lgtm

ricardozanini (approved):

PR #363, opened 2024-01-24
Fixes checks workflow by running apt-get update 

jordigilh:
Fixes checks workflow by running `apt-get update` before `apt-get install` and also adds flag `--no-install-recommends` to standardize with the other workflows since they share the same package installation call.

`--no-install-recommends` ensures that apt-get will only install the main dependencies, that's packages in the Depends field.

@ricardozanini @domhanak PTAL. This fixes the build workflow.

### DIFF
--- a/.github/workflows/check-container-builder.yml
+++ b/.github/workflows/check-container-builder.yml
@@ -58,7 +58,8 @@ jobs:
     steps:
       - name: Install packages
         run: |
-          sudo apt-get -y install \
+          sudo apt-get update && \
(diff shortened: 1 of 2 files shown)


domhanak (approved):
Thanks

ricardozanini:
This is not the only fix, there's more to add please see: https://github.com/apache/incubator-kie-kogito-serverless-operator/pull/362/commits/9d7e122f505b84627e48b94fbc6ec0e93261c8e3

ricardozanini:
Fixed in #365

PR #367, opened 2024-01-25
Break the current operator's configuration into custom and managed properties

### LINKED ISSUE #381: Break the current operator's configuration into custom and managed properties
### Description

*Summary*
Currently, the same {{ConfigMap}} holds managed (what we currently call "immutable") and custom properties defined by users.

A few application properties cannot be changed by users, being managed by the operator such as the HTTP port.

We should convert those properties to environment properties and set them to the deployment workflow. These env properties must be immutable and managed by the operator. Env properties override the application.properties mounted currently by the operator. See https://quarkus.io/guides/config-reference#configuration-sources

The current ConfigMap will be solely for user custom properties.

 

*Acceptance criteria:*

    After a successful deployment of the SonataFlow instance ABC, the ABC-props ConfigMap is still owned by this instance but not updated by the Sonata Flow operator
    All the immutable properties defined by the operator are mapped to env variables in the ABC Deployment owned by the SonataFlow instance (in the spec.template.spec.containers[0].env section)
    If any variable defined in the spec.podTemplate.container.env section of the SonataFlow instance match one of the immutable variable names, the value mapped in the ABC Deployment will be the one defined by the operator
        In this case, an Event of type Warning is created for the SonataFlow instance

### Implementation ideas

_No response_

dmartinol:
Closes #381 

- Added operator managed `ConfigMap` to store the managed properties (aka `immutable`)
  - Key `application-PROFILE.properties` contains the properties (`PROFILE` is either `prod` or `dev`)
- User properties are no more updated by the operator
  - Properties are defined by the key `application.properties`
- Any changes to one of the two property files implies a forced rollout of the workflow deployment
  - The checksum is calculated on the joined properties (user + managed)
- Since the managed properties are generated using the `SonataFlowPlatform` instance, all the creator and mutators have been extended to add a `platform` instance to the input parameters

More tests can be added after we agree on the overall approach.


ricardozanini:
We can add this link to the documentation: https://quarkus.io/guides/config-reference#default-profiles

ricardozanini:
@dmartinol can you please verify the e2e failures? 

```
Summarizing 1 Failure:
  [FAIL] SonataFlow Operator ensure that Operator and Operand(s) can run in restricted namespaces [It] should successfully deploy the orderprocessing workflow in devmode and verify if it's running
  /home/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.13.0/internal/node.go:463
```

Might be related with these changes.

workflowproj/workflowproj.go
```diff
@@ -38,6 +38,8 @@ import (
 	operatorapi "github.com/apache/incubator-kie-kogito-serverless-operator/api/v1alpha08"
 )
 
+// TODO: should we delete this whole file? seems used only the the associated test functions
```
ricardozanini, line 41: What do you mean? This is an external module that an actor can use to generate CRs based on SonataFlow projects. kn workflow uses it, for instance. 
But I'm curious to understand why you came to this conclusion. Maybe we need to communicate this better, as part of being part of the [README](https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/workflowproj/README.md).

controllers/profiles/common/ensurer.go
```diff
@@ -34,7 +34,7 @@ var _ ObjectEnsurer = &defaultObjectEnsurer{}
 var _ ObjectEnsurer = &noopObjectEnsurer{}
 
 type ObjectEnsurer interface {
-	Ensure(ctx context.Context, workflow *operatorapi.SonataFlow, visitors ...MutateVisitor) (client.Object, controllerutil.OperationResult, error)
+	Ensure(ctx context.Context, workflow *operatorapi.SonataFlow, platform *operatorapi.SonataFlowPlatform, visitors ...MutateVisitor) (client.Object, controllerutil.OperationResult, error)
```
ricardozanini, line 37: Instead, can't we have something like:

```go
type ObjectEnsurer interface {
	Ensure(ctx context.Context, workflow *operatorapi.SonataFlow, visitors ...MutateVisitor) (client.Object, controllerutil.OperationResult, error)
        WithPlatform(platform *operatorapi.SonataFlowPlatform)
}
```

Then in the ensurer that requires the platform you do `ensurer.WithPlatform(p).Ensure(...)`.

controllers/profiles/common/ensurer.go
```diff
@@ -62,10 +62,10 @@ type defaultObjectEnsurer struct {
 	creator ObjectCreator
 }
 
-func (d *defaultObjectEnsurer) Ensure(ctx context.Context, workflow *operatorapi.SonataFlow, visitors ...MutateVisitor) (client.Object, controllerutil.OperationResult, error) {
+func (d *defaultObjectEnsurer) Ensure(ctx context.Context, workflow *operatorapi.SonataFlow, platform *operatorapi.SonataFlowPlatform, visitors ...MutateVisitor) (client.Object, controllerutil.OperationResult, error) {
 	result := controllerutil.OperationResultNone
 
-	object, err := d.creator(workflow)
+	object, err := d.creator(workflow, platform)
```
ricardozanini, line 68: Here `defaultObjectEnsurer` can hold the `platform` and do a nil check to call the creator in the same fashion:

```go
if (d.platform == nil) {
    object, err := d.creator(workflow)
} else {
   object, err := d.withPlatform(platform).creator(workflow)
}
```

This way I think we can preserve the API since mutators and creators having a new unused parameter might be hard to maintain as you said in the chat.

controllers/profiles/dev/states_dev.go
```diff
@@ -62,23 +62,26 @@ func (e *ensureRunningWorkflowState) CanReconcile(workflow *operatorapi.SonataFl
...
+	flowDefCM, _, err := e.ensurers.definitionConfigMap.Ensure(ctx, workflow, pl, ensureWorkflowDefConfigMapMutator(workflow, pl))
+	if err != nil {
+		return ctrl.Result{Requeue: false}, objs, err
+	}
+	objs = append(objs, flowDefCM)
+	userPropsCM, _, err := e.ensurers.userPropsConfigMap.Ensure(ctx, workflow, pl)
+	if err != nil {
+		return ctrl.Result{Requeue: false}, objs, err
+	}
+	managedPropsCM, _, err := e.ensurers.managedPropsConfigMap.Ensure(ctx, workflow, pl, common.UserPropertiesMutateVisitor(ctx, e.StateSupport.Catalog, workflow, pl, userPropsCM.(*corev1.ConfigMap)))
```
ricardozanini, line 80: Does the `managedProps` include user properties too? Is this to calculate hash as you are describing in the PR? Can't we let the managed CM hold only manage props?

ricardozanini (commented):
I'll take a closer look later, but looks ok at a first glance. I let a proposal to avoid including the platform in every creator/mutator.

controllers/profiles/dev/states_dev.go
dmartinol: Yes, the idea is to hold only managed props, but the checksum is calculated on the aggregated string "user props, managed props", in order to detect any possible change

workflowproj/workflowproj.go
dmartinol: I was not aware of module this and I totally missed the `README`: now I understand that this is the tool used in `gen-manifest` utility. So, even if the functions in this file are not used in the operator, it must stay in the repo.

The reason why I bumped into this module is that it is performing some manipulation of the properties that now are no more needed: I will remove that part and also the generation of the `managed-props` ConfigMap that, I assume, will not be part of the `gen-manifest` output.
I will validate the manifests using the README, thanks for the pointer.

Please comment if you believe that the `gen-manifest` should not behave as I wrote above.

controllers/profiles/common/ensurer.go
dmartinol: I investigated this option and ATM I think it's not doable following your proposal, because the `creator` field is a function of type:
```go
type ObjectCreator func(workflow *operatorapi.SonataFlow) (client.Object, error)
```
and we cannot pass the `platform` reference to it, the signature is fixed.

What we can do, if you approve it, is to adopt an `interface` type instead, like:
```go
type ObjectCreator interface {
	Create(workflow *operatorapi.SonataFlow) (client.Object, error)
	WithPlatform(platform *operatorapi.SonataFlowPlatform) ObjectCreator
}
```
and then use it as:
```go
if (d.platform == nil) {
    object, err := d.creator.Create(workflow)
} else {
   object, err := d.creator.WithPlatform(platform).Create(workflow)
}
```

Of course, all the existing functions of type `ObjectCreator` have to be converted to `struct` with some basic implementation to hold the `platform` reference.

dmartinol:
> @dmartinol can you please verify the e2e failures?
> 
> ```
> Summarizing 1 Failure:
>   [FAIL] SonataFlow Operator ensure that Operator and Operand(s) can run in restricted namespaces [It] should successfully deploy the orderprocessing workflow in devmode and verify if it's running
>   /home/runner/go/pkg/mod/github.com/onsi/ginkgo/v2@v2.13.0/internal/node.go:463
> ```
> 
> Might be related with these changes.

I see the managed properties do not contain the "converted" discovery properties, so there was not mapping for this configuration:
```
mp.messaging.outgoing.kogito_outgoing_stream.url = ${kubernetes:services.v1/event-listener}
```
I'm investigating

controllers/profiles/dev/states_dev.go
ricardozanini: I see, but maybe we can do in a different function? Not sure if I followed the code exactly but it seems confusing having user and managed properties being merged in the function responsible by creating the CM. Maybe a `calculateHash(user, managed)` function?

controllers/profiles/common/ensurer.go
dmartinol: Added `ObjectEnsurerWithPlatform` to accomodate the latest proposal of using a specific creation function and ensurer only for selected resources

controllers/profiles/common/ensurer.go
dmartinol: Tried to accomodate the proposal in the latest commits, with few changes to properly pass-in the platform reference when it's needed. Working on the functional fixes now.

workflowproj/workflowproj.go
ricardozanini: Can you please just remove this line then?

dmartinol:
> We can add this link to the documentation: https://quarkus.io/guides/config-reference#default-profiles

can you remind me again the documentation repo?

controllers/profiles/dev/states_dev.go
dmartinol: please verify the latest commit.

dmartinol:
@ricardozanini apart from some E2E tests that are failing and I cannot replicate on my local environment, unit tests and manual verification seems to succeed. I will update the ADR in the meantime

ricardozanini:
> > We can add this link to the documentation: https://quarkus.io/guides/config-reference#default-profiles
> 
> can you remind me again the documentation repo?

Sure thing! https://github.com/apache/incubator-kie-kogito-docs/tree/main/serverlessworkflow/modules/ROOT

dmartinol:
> We can add this link to the documentation: https://quarkus.io/guides/config-reference#default-profiles

Adding to PR for doc repo

controllers/profiles/common/properties/discovery.go
```diff
@@ -99,6 +99,7 @@ func generateDiscoveryProperties(ctx context.Context, catalog discovery.ServiceC
 					mpProperty := generateMicroprofileServiceCatalogProperty(plainUri)
 					klog.V(log.I).Infof("Generating microprofile service catalog property %s=%s.", mpProperty, address)
 					result.MustSet(mpProperty, address)
+					result.MustSet(k, "http://localhost:8080")
```
ricardozanini, line 102: Are we adding `localhost` to discovered endpoints? That way we may lose track of it and we can't refresh the discovery if the service/endpoint changes.

controllers/profiles/common/mutate_visitors.go
```diff
@@ -105,32 +105,28 @@ func ServiceMutateVisitor(workflow *operatorapi.SonataFlow) MutateVisitor {
...
+			if !hasKey {
+				userProperties = ""
+			}
 			// In the future, if this needs change, instead we can receive an AppPropertyHandler in this mutator
 			props, err := properties.NewAppPropertyHandler(workflow, platform)
 			if err != nil {
 				return err
 			}
-			cm.Data[workflowproj.ApplicationPropertiesFileName] = props.WithUserProperties(cm.Data[workflowproj.ApplicationPropertiesFileName]).
+			managedProps.Data[workflowproj.GetManagedPropertiesFileName(workflow)] = props.WithUserProperties(userProperties).
```
ricardozanini, line 129: So the managed properties CM will have also user properties? Did I understand this correctly? Do we need both properties in the managed CM?

ricardozanini (commented):
Looks neat! I just have two questions.

ricardozanini:
I think you can rebase this PR to take latest Jordi changes to the tests.

controllers/profiles/common/mutate_visitors.go
dmartinol: no, the `userProps` are passed on to the property builder so that it can generate the discovery properties that, instead, are part of the managed properties.
see line 99 of `application.go`: `userProps` are not passed to the builder of the `managedProps`

ricardozanini: OH YEAH, you're right! Thanks for the clarification!

controllers/profiles/common/properties/discovery.go
dmartinol: this is a workaround to avoid the problem we saw yesterday.
if the `managed-props` does not override the value of the original property, the platform would anyway parse it and may raise a startup error like:
```
2024-01-29 16:15:52,397 orderprocessing-5f658b4c9d-wtmnp ERROR [io.quarkus.runtime.Application:-1] (Quarkus Main Thread) Failed to start application (with profile [dev]): java.lang.RuntimeException: Failed to start quarkus
...
Caused by: jakarta.enterprise.inject.spi.DeploymentException: java.lang.IllegalArgumentException: Unsupported scheme for the http connector in URL: services.v1/event-listener
...
```
(the property value being `${kubernetes:services.v1/event-listener}`)

If there is a more elegant workaround, we can give it a try, but for sure we cannot remove the original property from the user configuration as we did before.

dmartinol:
I'm touched by that green flag on the E2E checks 😂

controllers/profiles/common/properties/discovery.go
ricardozanini: @wmedvede we still have to assess this problem on the runtimes side. When we have a profile and standard properties in a discovery context.

dmartinol: I was wondering how I could create such discovery context, in fact. Our workflows are missing the `kogito-addons-quarkus-fabric8-kubernetes-service-catalog` extension, should we add it?
I've updated one of our application to define an openAPI endpoint as:
```
mta_ui = ${kubernetes:services.v1/openshift-mta/mta-ui?port=ui}
quarkus.rest-client.mta_json.url = ${mta_ui}/hub
```
which is then mapped into the managed properties:
```
mta_ui = http://localhost:8080
org.kie.kogito.addons.discovery.kubernetes\:services.v1\/openshift-mta\/mta-ui?port\=ui = http://mta-ui.openshift-mta.svc:8080
```
But this I believe needs that particular extension to be resolved, correct?

dmartinol: I think I solved my specific use case by changing to:
```
result.MustSet(k, address)
```
that leads to these two managed properties instead:
```
org.kie.kogito.addons.discovery.kubernetes\:services.v1\/openshift-mta\/mta-ui?port\=ui = http://mta-ui.openshift-mta.svc:8080
mta_ui = http://mta-ui.openshift-mta.svc:8080
```
where the first one is not in use because of the missed extension (my guess)

ricardozanini: @dmartinol we ditched the `addons-quarkus-fabric8-kubernetes-service-catalog` on purpose since we implemented the discovery in the operator, to not depend on Fabric8 and giving the pod unneeded permissions.

That said, the underlying runtime now relies on smallrye mpconfig to do the "discovery", which is this props dictionary we mount in the CM. So now we use the [microprofile-config catalog](https://github.com/apache/incubator-kie-kogito-runtimes/tree/main/quarkus/addons/microprofile-config-service-catalog) instead.

So in your example, our managed props take over `mta_ui`? Since we will have:

User Properties:

```properties
mta_ui = ${kubernetes:services.v1/openshift-mta/mta-ui?port=ui}
quarkus.rest-client.mta_json.url = ${mta_ui}/hub
```
Managed:
```
org.kie.kogito.addons.discovery.kubernetes\:services.v1\/openshift-mta\/mta-ui?port\=ui = http://mta-ui.openshift-mta.svc:8080
mta_ui = http://mta-ui.openshift-mta.svc:8080
```

Hence, `mta_ui` must override what's set on the user profile. It will probably work since the profile takes precedence, right?

That will work considering we keep updating `mta_ui` on every reconciliation if a user changes their service endpoint.

dmartinol: @ricardozanini I finally understood that `we implemented the discovery in the operator`, as you said. After reading the nice introduction to the [service discovery add-on](https://blog.kie.org/2022/10/kogito-service-discovery-for-workflows.html) by Filippe, I believe this is the way to go.

I will raise an updated commit after a quick validation.

wmedvede:
@dmartinol looks like we need some love here to fix merge conflicts

dmartinol:
> @dmartinol looks like we need some love here to fix merge conflicts

and also updated some props, but I need to re-run my manual tests, I'm concerned that latest commits broke anything.
BTW: about properties-related tests, it doesn't make much sense that we only check the number of generated props, but at this point I cannot add a check for each expected properties, as this set is changing day by day. So, I just changed the expected number.

controllers/profiles/common/properties/discovery.go
ricardozanini: Thank you!

controllers/profiles/common/mutate_visitors.go
wmedvede: I think this code for people that comes from the previous version is a bit confusing in the first read, because before, the user properties where combined with the properties we where generating. And now, with the internal refactoring you did, the user properties are  being used as a kind of "candidates for doing the service discovery", but won't be included in the produced properties set.

Maybe, now that we are going to keep to properties living in parallel, the original application.properties, plus the new application-pod.properties generated by this handler, we can rename to something like ManagedAppPropertiesHandler (or a better name)  to distinguish the new situation.

Just thinking aloud.

controllers/profiles/common/properties/application.go
```diff
@@ -75,36 +75,34 @@ func (a *appPropertyHandler) WithServiceDiscovery(ctx context.Context, catalog d
 }
 
 func (a *appPropertyHandler) Build() string {
-	var props *properties.Properties
+	var userProps *properties.Properties
```
wmedvede, line 78: Related to the comment I did before, this ApplicationPropertyHandler has changed the original semantic. Maybe we should rename and update a bit the documentation of the function NewAppPropertyHandler.

controllers/profiles/common/mutate_visitors.go
```diff
@@ -105,32 +105,28 @@ func ServiceMutateVisitor(workflow *operatorapi.SonataFlow) MutateVisitor {
 	}
 }
 
-func WorkflowPropertiesMutateVisitor(ctx context.Context, catalog discovery.ServiceCatalog,
-	workflow *operatorapi.SonataFlow, platform *operatorapi.SonataFlowPlatform) MutateVisitor {
+func UserPropertiesMutateVisitor(ctx context.Context, catalog discovery.ServiceCatalog,
```
wmedvede, line 108: Maybe we should name this ManagedPropertiesMutateVisitor instead, wdyt?

controllers/profiles/common/properties/discovery.go
dmartinol: Latest commits include the above change and I manually validated it on an OCP deployment with the above use case.

controllers/profiles/common/test/common.go
```diff
@@ -0,0 +1,84 @@
...
+
+package common_test
+
+import (
+	"context"
+
+	"github.com/apache/incubator-kie-kogito-serverless-operator/controllers/discovery"
+)
+
+const (
```
wmedvede, line 28: This file was originally generated to leave in the context of the properties tests package, why do we need to move it outside it?

The question is because as we move to a common package the trend is that it'll degenerate and become a sort of all in one where people start putting the stuff their need for whatever test, etc.

wmedvede (commented):
I have left some minor comments, good job!

controllers/profiles/common/ensurer.go
```diff
@@ -36,6 +36,9 @@ var _ ObjectEnsurer = &noopObjectEnsurer{}
 type ObjectEnsurer interface {
 	Ensure(ctx context.Context, workflow *operatorapi.SonataFlow, visitors ...MutateVisitor) (client.Object, controllerutil.OperationResult, error)
 }
+type ObjectEnsurerWithPlatform interface {
```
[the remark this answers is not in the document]
ricardozanini: @wmedvede not all ensurers request platform. It's bad adding this unused parameter in some cases just because the API requires it. That's why we opted to add a new interface.

wmedvede:
@dmartinol @ricardozanini @domhanak some positive feedback for the record.

I have executed **usecase-oc** https://github.com/flows-examples/techpreview2/blob/f39a25e626e217dabf7b69237641ca1e05375c6a/README.md?plain=1#L739

And I can see this results:

**Managed and normal properties are created** 
Screenshot from 2024-02-02 17-18-38

**I've added a service discovery property manually in the callbackstate-props**

Screenshot from 2024-02-02 17-18-59

Then I went to the managed properties and I can see that the service discovery was produced and the corresponding property is in the managed set. The other managed properties are still there. good!


Screenshot from 2024-02-02 17-19-21

- The workflow  is running well I could create instances and see them in the DI, good!

- When I add properties manually to the user properties I can then see the workflow rollingout. good!


- In the workflow deployment environment I can't see any property
But maybe we can start moving things there in a follow-up PR. If I understood well, we should start to see there this kind of props
    quarkus.http.port = 8080
    quarkus.http.host = 0.0.0.0
as env vars.

But we can think about this in follow-up PR

image


good job @dmartinol !!!!

wmedvede (approved):
@dmartinol @ricardozanini approving from my side after doing some local testing since my comments are minor suggestions.

dmartinol:
> @dmartinol @ricardozanini @domhanak some positive feedback for the record.
> 
> * In the workflow deployment environment I can't see any property
>   But maybe we can start moving things there in a follow-up PR. If I understood well, we should start to see there this kind of props
>   quarkus.http.port = 8080
>   quarkus.http.host = 0.0.0.0
>   as env vars.
> 
> But we can think about this in follow-up PR
> 
Thanks for deep investigation and testing.
Wrt the deployment environment, we changed our initial design because this option doesn't work well with property names including "-", see this issue [smallrye-reactive-messaging: incompatibility with configuration through environment variables](https://github.com/quarkusio/quarkus/issues/30106). E.g., `mp.messaging.outgoing.kogito-job-service-job-request-events.connector` cannot be managed that way.

If you aim to model as env variables only the properties not affected by that issue, this is an option but at the same time we should keep the managed properties file for all of the others, so we'd have 3 managed source of configuration values.

controllers/profiles/common/test/common.go
dmartinol: I also didn't like this move but it was needed because, in one of the commits (long history), those functions and constants were used by test functions in different packages, which created a sort of circular dependency. maybe now I can put it back into the original package, let me investigate. Thanks for raising this

dmartinol: Good catch @wmedvede: I did not realize that some of the recent rebases have restored the original code, so that `common_test` package was not used at all, so I just removed it. Thanks!

controllers/profiles/common/mutate_visitors.go
dmartinol: Pls see the updates on the latest commit

wmedvede:
@dmartinol @ricardozanini @domhanak 

I have found a small non blocking issue here.

**When playing to add service discovery properties in the callbackstatetimeouts-prop I define this:**

Screenshot from 2024-02-02 18-09-43

After the managed properties calculation was produced in the Reconcile, I can see that the service discovery properties where created, good, but, user.property3 and user.property4 was copied to the managed properties too with a sort of calculated value.
This didn't happen before. It looks like a property expansion is enabled now during the properties calculation or something similar.

 
Screenshot from 2024-02-02 18-10-05

wmedvede:
> > @dmartinol @ricardozanini @domhanak some positive feedback for the record.
> > 
> > * In the workflow deployment environment I can't see any property
> >   But maybe we can start moving things there in a follow-up PR. If I understood well, we should start to see there this kind of props
> >   quarkus.http.port = 8080
> >   quarkus.http.host = 0.0.0.0
> >   as env vars.
> > 
> > But we can think about this in follow-up PR
> 
> Thanks for deep investigation and testing. Wrt the deployment environment, we changed our initial design because this option doesn't work well with property names including "-", see this issue [smallrye-reactive-messaging: incompatibility with configuration through environment variables](https://github.com/quarkusio/quarkus/issues/30106). E.g., `mp.messaging.outgoing.kogito-job-service-job-request-events.connector` cannot be managed that way.
> 
> If you aim to model as env variables only the properties not affected by that issue, this is an option but at the same time we should keep the managed properties file for all of the others, so we'd have 3 managed source of configuration values.

ok, we can see this in a future step.

ricardozanini:
@wmedvede I'd rather keep all managed properties in the CM. One place to look for the managed properties by the operator can facilitate admin tasks.

Regarding the bug you've found, this is intentional to override the ones by the user. Otherwise, Quarkus will try to read them and parse. Since `kubernetes:` is not a valid schema, it will fail.

Actually, by chance, @dmartinol found a way to remove our `microprofile-config` add-on. We don't need it anymore since it's already doing the expansion on the operator side. So in your example, only these will do:

> application-dev.properties (Managed)
```properties
user.property3=<discovered URL>
user.property4=<discovered URL>
```

> application.properties (User)
```properties
user.property3=<desired Kubernetes endpoint>
user.property4=<desired Kubernetes endpoint>
```

We can ditch the jar and live happily.

wmedvede:
> We can ditch the jar and live happily.

np from my side, the only thing is that I personally don't know how these properties has ended up in the managed property set, and also with the calculated values

user.property3=<desired Kubernetes endpoint>
user.property4=<desired Kubernetes endpoint>

Then, regarding the microprofile like generated properties and the removal of that addon, I think we need to investigate a bit, because when we configure knative-serving operations like:

      "operation": "knative:services.v1.serving.knative.dev/default/serverless-workflow-greeting-quarkus?path=/arrayFunction"

To make them work, if I'm not wrong we need the microprofile addon, need to double check on monday.

ricardozanini:
@dmartinol can you please try rebasing? Since we've merged the label PR, I think this line in the e2e is causing troubles:

`kubectl wait pod -A -l control-plane=sonataflow-operator --for condition=Ready`

Your fork still has the old label: https://github.com/dmartinol/incubator-kie-kogito-serverless-operator/blob/SRVLOGIC-195-props/.github/workflows/e2e.yml#L74

ricardozanini:
> need to double check on Monday.

@wmedvede no worries. The problem is that our addon should consider the property in the default profile, which is not. That's why we had to override in the managed CM.

domhanak (approved):

ricardozanini:
@dmartinol once you rebase, CI is green, I'll merge this.

PR #374, opened 2024-01-26
kie-issues#821: Kogito Serverless Operator weekly job

rodrigonull:
ISSUE: https://github.com/apache/incubator-kie-issues/issues/821

This PR introduces a new weekly Jenkins job for Kogito Serverless Operator that updates the project version to 999.0.0-YYYYMMDD-snapshot locally, build it and deploy the image to quay.io

The weekly job also creates a new tag in the repo with this format: 999.0.0-YYYYMMDD.

### DIFF
--- a/.ci/jenkins/Jenkinsfile.weekly.deploy
+++ b/.ci/jenkins/Jenkinsfile.weekly.deploy
@@ -1,0 +1,252 @@
+@Library('jenkins-pipeline-shared-libraries')_
+
+helper = null
+
+pipeline {
+    agent {
+        docker {
+            image env.AGENT_DOCKER_BUILDER_IMAGE
+            args env.AGENT_DOCKER_BUILDER_ARGS
+        }
+    }
+
+    options {
+        timeout(time: 10, unit: 'HOURS')
+        timestamps()
+    }
+
+    environment {
+        KOGITO_CI_EMAIL_TO = credentials("${JENKINS_EMAIL_CREDS_ID}")
+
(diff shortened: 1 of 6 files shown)


ricardozanini (commented):
I think we can use 10.0.0 since we should follow KIE release train. WDYT?

rodrigonull:
> I think we can use 10.0.0 since we should follow KIE release train. WDYT?

Perhaps we can set the version to 999-snapshot as discussed here: https://github.com/apache/incubator-kie-issues/issues/763? WDYT @ricardozanini @domhanak

rodrigonull:
> @rodrigonull is the jenkinsfile a "close" copy of an existing one? Wouldn't reuse of a nightly one be possible with some tweaks? Also please can you check that the triggering part is in place? Or please point me to where it's defined.

Yes, this Jenkinfile is very similar to the nightly one. I've decided to create a new jenkinfile due to my lack o experience with the pipelines framework and to avoid breaking existing jobs. There are some differences between the nightly and the weekly jobs as described bellow:

for weekly jobs we need to:
Checkout the code from a specific datetime
Update the project version to a specific version (999-YYYYMMDD pattern)
Create and publish a tag in the repository

I think at some point would be nice to refactor this work and create a new job type for weekly jobs and update the nightly jenkinsfiles to support weekly jobs. WDYT?

The PR that adds the triggering is this one: https://github.com/apache/incubator-kie-kogito-pipelines/pull/1177

PR #383, opened 2024-02-04
Fixes kind create cluster script that contained a hardcoded IP

jordigilh:
Removes a hardcoded IP that defined the directory inside `/etc/containerd` for the registry configuration. When the IPs mismatched it caused a failure to pull or push images from the insecure registry via kaniko builder with Kind.   The correct approach is to use the IP from the docker network configuration for the registry container for both the directory name and the configuration contained within.

@wmedvede @ricardozanini PTAL.


ricardozanini (approved):
Thank you!

ricardozanini:
Can you please rebase your fork? 🙏
This is the reason why e2e failed.

domhanak (approved):

PR #414, opened 2024-02-29
Fix Issue #319 - Rename prod profile to preview

### LINKED ISSUE #319: Rename "prod" profile to "preview"
### Description

Calling "prod" profile might confuse users into thinking that the operator can deliver an immutable image ready for production use cases. While this can be true in simplistic scenarios, real use cases require much more than the simple embedded build platform that we currently ship with the operator.

On Kubernetes we rely on Kaniko, on OpenShift on a simple `BuildConfig` object. This setup is more for a "preview" or staging use cases where user wants a glimpse of what it will look like a lighter image and immutable workflow on their topology.

Real use cases must rely on external build tooling. An example can be seen in this repo: https://github.com/flows-examples/gitops

This proposal is to change the `prod` profile naming and internal modules to `preview`. The profile capable to analyze and deploy an external built image we will call internally to `gitops` and it's activated as long as there's a `.spec.podTemplate.flowContainer.image` valid attribute set.

### Implementation ideas

Simply renaming and refactoring. No new features to add. Also, the guides must be updated accordingly.

ricardozanini:
Fixes #319 

**Checklist**

<details>
<summary>
</summary>

</details>


workflowproj/operator.go
```diff
@@ -75,9 +75,9 @@ func GetWorkflowManagedPropertiesConfigMapName(workflow *operatorapi.SonataFlow)
 
 // GetManagedPropertiesFileName gets the default ConfigMap name that holds the managed application property for the given workflow
 func GetManagedPropertiesFileName(workflow *operatorapi.SonataFlow) string {
-	profile := metadata.ProdProfile
+	profile := metadata.QuarkusProdProfile
```
wmedvede, line 78: So here we have kind of naming separation, by one side we might be executing the SFW "Preview" profile, but, we create an application-prod.prperties because the quarkus profiling?

ricardozanini: Yes, `prod` is the default Quarkus profiling name: https://quarkus.io/guides/config-reference#default-profiles

ricardozanini: SonataFlow Preview profile uses Quarkus prod profile. We are not mimicking/exposing Quarkus profiles.

wmedvede: yep, I know that "prod" is the default quarkus profile. My question was more related to the fact that when users works with the "Dev" profile, they'll see "application-dev", but, when they work with "Preview" profile, they'll see "application-prod".  Np, but I think this must be clarified in the docs maybe.

wmedvede (approved):

workflowproj/operator.go
ricardozanini: will do thanks!

ricardozanini:
@jakubschwan can you please review? It is a must since it will force other PRs to rebase.

PR #415, opened 2024-03-01
[KOGITO-9973] Add `SonataFlow.Status.Services` field and improve workflow properties ConfigMap reconciliation

tchughesiv:
This PR adds supporting services to the `SonataFlow` status. With this change, we'll start using the `SonataFlow.Status.Services` when generating workflow properties ConfigMaps. This is to ensure `SonataFlow.Status.Services` matches & controls what is set in the managed application.properties ConfigMap. 

This PR also ensures that all supporting services changes, whether cluster-wide or local to the workflow, are reflected via ConfigMap reconciliation... which automatically triggers the workflow deployment to rollout and use the new properties.

An example of the new status fields -
```yaml
apiVersion: sonataflow.org/v1alpha08
kind: SonataFlow
metadata:
  name: workflow
  namespace: <namespace>
spec:
  ...
status:
  ...
  services:
    dataIndexRef:
      url: 'http://example-data-index-service.<namespace>'
    jobServiceRef:
      url: 'http://example-jobs-service.<namespace>'
```

To reflect which supporting services are being used by each workflow so the user has a clearer understanding.
To ensure supporting services settings/changes are reconciled to workflows cluster-wide. This is critical. For example, we don't want workflows attempting to communicate against services which have been disabled or deleted. The workflows should always reflect the current state of the supporting services.

**Checklist**

<details>
<summary>
</summary>

</details>


controllers/profiles/common/reconciler.go
```diff
@@ -46,7 +48,12 @@ type StateSupport struct {
 // PerformStatusUpdate updates the SonataFlow Status conditions
 func (s StateSupport) PerformStatusUpdate(ctx context.Context, workflow *operatorapi.SonataFlow) (bool, error) {
 	var err error
+	pl, err := platform.GetActivePlatform(ctx, s.C, workflow.Namespace)
```
ricardozanini, line 51: Don't we already have the platform in the `StateSupport`?

ricardozanini (commented):
Looks great! Just a minor comment. I remember we setting the platform to the controller states back in a Danielle or Jordi's PR.

controllers/profiles/common/reconciler.go
tchughesiv: hmm... i'm not seeing it
https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/76ef90247c695ea6973e7887c794b03a1541ac63/controllers/profiles/common/reconciler.go#L41-L47

tchughesiv:
i need to modify the e2e tests before we merge to account for this change.

tchughesiv:
ok, e2e tests have been added and should be working

controllers/profiles/common/reconciler.go
ricardozanini: I'm wondering if we can reuse the platform we already get in the states/reconciliation. But it's a nit; I think the PR is fine. 👍

ricardozanini (approved):

ricardozanini:
@wmedvede @jakubschwan can you please review? @tchughesiv will have to jump into another task.

controllers/platform/services/properties.go
```diff
@@ -185,19 +186,21 @@ func GenerateDataIndexWorkflowProperties(workflow *operatorapi.SonataFlow, platf
...
 	js := NewJobServiceHandler(platform)
-	if workflow != nil && !profiles.IsDevProfile(workflow) && js.IsServiceEnabled() {
-		if workflowdef.HasTimeouts(workflow) {
-			props.Set(constants.KogitoJobServiceHealthCheckEnabled, "true")
-		}
-		p, err := js.GenerateWorkflowProperties()
-		if err != nil {
-			return nil, err
+	if !profiles.IsDevProfile(workflow) && workflow != nil && workflow.Status.Services != nil && workflow.Status.Services.JobServiceRef != nil {
+		serviceBaseUrl := workflow.Status.Services.JobServiceRef.Url
```
wmedvede, line 191: I see something that I believe more difficult to follow now, or potentially error prone. Let me explain:

**Before:**
    The jobs-service url was calculated by the JobServiceHandler, and there, it was derived somehow directly from that service, or, from the cluster platform if any.

**Now:**
    Get get the jobs-service url from the  workflow.Status.Services.JobServiceRef, BUT, the workflow is normally what is being deployed at this point, and that somehow "caused" the execution of all this cycle. So we configure some WF properties from values from the WF itself.

I am correct?


Same reasoning apply for the DI

controllers/platform/services/properties.go
```diff
@@ -159,24 +159,25 @@ func generateReactiveURL(postgresSpec *operatorapi.PersistencePostgreSQL, schema
...
+			props.Set(constants.KogitoProcessDefinitionsEventsErrorsEnabled, "true")
+			props.Set(constants.KogitoProcessInstancesEventsEnabled, "true")
+			props.Set(constants.KogitoDataIndexHealthCheckEnabled, "true")
+			props.Set(constants.KogitoDataIndexURL, serviceBaseUrl)
+			props.Set(constants.KogitoProcessDefinitionsEventsURL, serviceBaseUrl+constants.KogitoProcessDefinitionsEventsPath)
+			props.Set(constants.KogitoProcessInstancesEventsURL, serviceBaseUrl+constants.KogitoProcessInstancesEventsPath)
 		}
-		props.Merge(p)
+	} else {
+		props.Set(constants.KogitoProcessDefinitionsEventsEnabled, "false")
```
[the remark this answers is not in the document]
tchughesiv: it'll work either way, I'll change it.

controllers/platform/services/properties.go
```diff
@@ -185,19 +186,21 @@ func GenerateDataIndexWorkflowProperties(workflow *operatorapi.SonataFlow, platf
 // Never nil.
 func GenerateJobServiceWorkflowProperties(workflow *operatorapi.SonataFlow, platform *operatorapi.SonataFlowPlatform) (*properties.Properties, error) {
 	props := properties.NewProperties()
-	props.Set(constants.JobServiceRequestEventsConnector, constants.QuarkusHTTP)
```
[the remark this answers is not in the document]
tchughesiv: done

controllers/platform/services/properties.go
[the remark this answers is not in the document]
tchughesiv: done

controllers/platform/services/properties.go
tchughesiv: no, we are still using the service handlers to set the urls. There's a new method called `SetServiceUrlsInWorkflowStatus` that handles this in the workflow(s) status -
https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/2325390963e643afd9868174e9818d7dafde4317/controllers/platform/services/services.go#L432-L439

what HAS changed is... during a workflow's configmap properties creation, instead of calling the handler directly, we're using the service url directly from the workflow's status... which has already been properly set by the service handlers.

this is actually LESS error prone because it ensures the accuracy of some things, which, in `main` today can be a bit nebulous -

1. this ensures that a workflow's application properties match the correct service urls which are set the status of a workflow. currently a user would have to figure this out for themselves by either checking the config map, or the platform object.
2. this ensures that application properties config maps are properly updated in real-time when service changes are made in the platform/clusterplatform... which, in turn, forces the workflow pod to restart and these new settings to take affect. today, a user may have to remove the existing configmap to force proper properties... which is problematic because they may not know this is required and a workflow may be attempting to send to a non-existent (or incorrect) service.

tchughesiv: its important to keep in mind that, workflows are re-reconciled anytime a related platform object is changed -
https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/fff6361bde07b566a22bc24ec7be84506e7b98ae/controllers/sonataflow_controller.go#L174-L181

this ensures that service-related changes at the platform level will result in a workflow.status change... which will result in a configmap change and workflow pod restart. this has all been thoroughly tested.

tchughesiv: one more note on this. this is the final part of a long planned, multi-tasked, implementation of supporting services in the operator. `workflow.status.services` has not existed until now, so we previously had no choice but to call the service handlers directly during workflow properties creation. with this PR, we'll have reached the intended "final" implementation.

wmedvede: Ok, many thanks for the explanation.
The part that wasn't clear to me was "during a workflow's configmap properties creation, instead of calling the handler directly, we're using the service url directly from **the workflow's status... which has already been properly set by the service handlers.**"

wmedvede (approved):
good job!

PR #416, opened 2024-03-05
Issue #312 - Create a new controllers config package for handling global configuration options for admins

### LINKED ISSUE #312: Avoid hardcoding resources elements required by the operator
### Description

By default, when deploying the operator via bundle, OLM will create a configMap named `sonataflow-operator-builder-config`:

https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/6c70c4cbfa805aad96e491e7f2e0699c2214b0ed/controllers/builder/config.go#L35

Not only this but there are also other places where the operator sets a "default" value that can possibly be overridden by an admin when installing the operator on different platforms.

We may introduce a ConfigMap to carry all this information and rely on hardcoded constants only when this info is not available. This configMap must be deployed with the manager within the same namespace and it's name should be bound to an env var on the manager's deployment.

### Implementation ideas

1. Add the hardcoded information in the Manager's ConfigMap
2. Have the Manager's ConfigMap name bind into an env var on its deployment.

ricardozanini:
Closes #312

**Checklist**

<details>
<summary>
</summary>

</details>


controllers/platform/services/services.go
```diff
@@ -265,6 +269,9 @@ func (j JobServiceHandler) GetContainerName() string {
 }
 
 func (j JobServiceHandler) GetServiceImageName(persistenceName string) string {
+	if len(cfg.GetCfg().JobsServiceImageTag) > 0 {
+		return cfg.GetCfg().JobsServiceImageTag
```
wmedvede, line 273: Not sure if this is good, or at least it's not aligned to what we had. Let me explain:
In order to get the proper jobs-service image, right now, two variables are considered, the "persitenceName"  and the "tag" to use.
For example, if no persistence is configured, "ephemeral" persistence will be considered, and thus, the returned image will be  kogito-jobs-service-ephemeral-nightly:latest.
On the other hand, if postgresql is configured, then, kogito-jobs-service-postgresql-nightly:latest will be used instead.

So, the image might vary depending on the persistence type we are using.

With your change, only one image JobsServiceImageTag is admitted.  Is that correct?

controllers/platform/services/services.go
```diff
@@ -101,6 +102,9 @@ func (d DataIndexHandler) GetContainerName() string {
 }
 
 func (d DataIndexHandler) GetServiceImageName(persistenceName string) string {
+	if len(cfg.GetCfg().DataIndexImageTag) > 0 {
```
wmedvede, line 105: same comment as for the jobs-service

controllers/platform/services/services.go
ricardozanini: Oh yeah, I forgot about that, I'll add two more configuration options there to solve this.
Thank you!

ricardozanini:
@wmedvede can you please review again?

wmedvede (approved):

ricardozanini:
@domhanak @jakubschwan ptal?

bundle/manifests/sonataflow-operator-controllers-config_v1_configmap.yaml
```diff
@@ -0,0 +1,34 @@
...
+    jobsServiceEphemeralImageTag: ""
+    # The Data Index image to use, if empty the operator will use the default Apache Community one based on the current operator's version
+    dataIndexPostgreSQLImageTag: ""
+    dataIndexEphemeralTag: ""
+    # SonataFlow base builder image used in the internal Dockerfile to build workflow applications in preview profile
+    # Order of precedence is:
+    # 1. SonataFlowPlatform in the given namespace
+    # 2. This configuration
+    # 3. The FROM in the Dockerfile in the operator's namespace "sonataflow-operator-builder-config" configMap.
+    # If 1 or 2, the FROM tag will be replaced by the tag se there.
```
domhanak, line 24: Could you please clarify `se` ?

bundle/manifests/sonataflow-operator-controllers-config_v1_configmap.yaml
```diff
@@ -0,0 +1,34 @@
...
+    jobsServicePostgreSQLImageTag: ""
+    jobsServiceEphemeralImageTag: ""
+    # The Data Index image to use, if empty the operator will use the default Apache Community one based on the current operator's version
+    dataIndexPostgreSQLImageTag: ""
+    dataIndexEphemeralTag: ""
+    # SonataFlow base builder image used in the internal Dockerfile to build workflow applications in preview profile
+    # Order of precedence is:
+    # 1. SonataFlowPlatform in the given namespace
+    # 2. This configuration
+    # 3. The FROM in the Dockerfile in the operator's namespace "sonataflow-operator-builder-config" configMap.
```
domhanak, line 23: `sonataflow-operator-builder-config` shouldn't this be parametrized?

domhanak (approved):
Changes look good to me, E2E tests work. I did not do any exploratory tests

@jakubschwan please take a look as you are better SME for this.

bundle/manifests/sonataflow-operator-controllers-config_v1_configmap.yaml
davidesalerno: I guess is probably `set`

jakubschwan (approved):

PR #435, opened 2024-03-22
Issue-434: Forcing pulling images from BC when latest is set

### LINKED ISSUE #434: Set `PullPolicy` to default on our deployments and builds to avoid using old images when tag is latest
### Description

See more: https://kubernetes.io/docs/concepts/containers/images/#updating-images

We should avoid setting `PullPolicy` in our definitions since the Kubernetes default can handle all sorts of scenarios (see https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting)

In summary, if the tag is set to `:latest`, k8s will set the`PullPolicy` to "Always". Otherwise, to "IfNotPresent".

If we set it to `IfNotPresent` as it is currently, when the tag is `:latest` and it's present in the cache, it won't pull the registry.

For the `BuildConfig` behavior with Dockerfiles FROM clause, we need more investigation. Looks like the BC is not pulling the image from the registry if the image is already present locally even if the tag is set to `:latest`.

### Implementation ideas

_No response_

ricardozanini:
In this PR, we turn a flag in the internally managed OpenShift build to force pull the images in the builder Dockerfile to refresh the `latest` tag.

Together with #416 we should be able to have a more robust and flexible approach to deal with image tags.

Fixes #434 

**Checklist**

<details>
<summary>
</summary>

</details>


wmedvede (approved):

ricardozanini:
@jakubschwan @domhanak can you please verify?

jakubschwan (approved):

PR #447, opened 2024-04-16
Fix #385 - Introduce Knative deployments for workflows 

### LINKED ISSUE #385: Opt-in deploy SonataFlow as Knative Serving service instead of k8s Deployments
### Description

Currently, the operator deploys workflows as regular Kubernetes Deployments. In some use cases, deploying a workflow as a Knative Serving resource makes sense.

- Assess the possibility of deploying workflows as Knative Serving services by default if Knative is installed in the cluster
- Add the possibility for users to opt-in to deploying as a ksvc or k8s deployment by labeling the CR.
- Once labeled and deployed, this attribute must be immutable. We may assess the possibility of changing this field later in the field
- Initially, revisions handling is out of scope, the operator won't reconcile this field so that users can change it based on their requirements. An assessment of revisions in the platform can be done later

### Implementation ideas

_No response_

ricardozanini:
Closes #385 

**Checklist**

<details>
<summary>
</summary>

</details>


ricardozanini:
@domhanak this is a good candidate for BDD. I'll do more manual testing tomorrow, but I can't add a Knative installation to KIND this time since it will add more overhead to our (already) resource-hungry e2e.

One more thing: I think the services, persistence, and knative support should be added before starting any ITs and reused across test cases. Currently, the e2e creates and destroys DBs and supporting services deployment every cycle, which is far from ideal.

domhanak:
@ricardozanini +1

I am working on baseline configurations to have for testing, and then each baseline has scenarios on top, different confguration sup. services config, no services, ephemeral etc.

api/metadata/annotations.go
```diff
@@ -73,3 +73,30 @@ const (
 	// Ideally used in production use cases
 	GitOpsProfile ProfileType = "gitops"
 )
+
+const (
+	DefaultProfile = PreviewProfile
+)
+
+var supportedProfiles = map[ProfileType]ProfileType{DevProfile: DevProfile, PreviewProfile: PreviewProfile, GitOpsProfile: GitOpsProfile}
```
jakubschwan, line 81: Add note that deprecated profile is not supported
```suggestion
// deprecated prod profile is deprecate and not supported, use preview profile
var supportedProfiles = map[ProfileType]ProfileType{DevProfile: DevProfile, PreviewProfile: PreviewProfile, GitOpsProfile: GitOpsProfile}
```
Not sure how exactly we handle deprecated profiles, but it would be good to add there some note. In case we do not need deprecated profile anymore, we can also remove it for next release.

api/metadata/annotations_test.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+	tests := []struct {
+		name string
+		args args
+		want ProfileType
+	}{
+		{"Empty Annotations", args{annotation: nil}, DefaultProfile},
+		{"Non-existent Profile", args{annotation: map[string]string{Profile: "IDontExist"}}, DefaultProfile},
+		{"Regular Annotation", args{annotation: map[string]string{Profile: GitOpsProfile.String()}}, GitOpsProfile},
+		{"Deprecated Annotation", args{annotation: map[string]string{Profile: ProdProfile.String()}}, DefaultProfile},
```
jakubschwan, line 33: Add Dev profile to tests
```suggestion
		{"Empty Annotations", args{annotation: nil}, DefaultProfile},
		{"Non-existent Profile", args{annotation: map[string]string{Profile: "IDontExist"}}, DefaultProfile},
		{"Regular Annotation", args{annotation: map[string]string{Profile: GitOpsProfile.String()}}, GitOpsProfile},
		{"Deprecated Annotation", args{annotation: map[string]string{Profile: ProdProfile.String()}}, DefaultProfile},
		{"Dev Annotation", args{annotation: map[string]string{Profile: DevProfile.String()}}, DevProfile},
```

controllers/profiles/gitops/profile_gitops_test.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 func Test_Reconciler_ProdOps(t *testing.T) {
-	workflow := test.GetBaseSonataFlowWithProdOpsProfile(t.Name())
+	workflow := test.GetBaseSonataFlowWithGitOpsProfile(t.Name())
```
jakubschwan, line 32: Is this change correct, as instead of prod profile should be used preview?

ricardozanini: ops, yup, should be Preview. It's just the naming, thanks @jakubschwan!

wmedvede (commented):
I have deployed a workflow like this:

```
apiVersion: sonataflow.org/v1alpha08
kind: SonataFlow
metadata:
  name: callbackstatetimeouts
  annotations:
    sonataflow.org/description: Callback State Timeouts Example k8s
    sonataflow.org/version: 0.0.1
    sonataflow.org/profile: preview
spec:
  podTemplate:
    deploymentModel: knative
  flow
    bla, bla, bla
```

And I get a regular kubernetes deployment and service, not a knative Service.

ricardozanini:
@wmedvede many thanks, I've updated the roles and permissions!

wmedvede:
By doing a local manipulation of the sonataflow-operator-manager-role ClusterRole, and patching manually to something like this:

https://gist.github.com/wmedvede/6c2bd803e1b03f24a3101fa2066c6788

I was able to deploy the greeting workflow, the builder executed well, and the knative service was created.
I'll to do now a similar deployment by using  a sonataflow with DI and JS installed.
 
image

wmedvede:
If we go back the workflow to the "normal mode" by removing the
```

  podTemplate:
      deploymentModel: knative
```

we get the kubernetes Deployment and Service as expected.

ricardozanini:
@wmedvede Actually we should delete the workflow and recreate it. But we can explore this in a new issue.

wmedvede:
Using now the callbackstatetimeouts with DI and JS deployed, we can see the knative service created as expected

image

wmedvede:
And on the DI we can see that the process instance was properly registred and also a corresponding job was executed to wake-up the service. In more exhaustive test we must verify that the service was properly waked up by the jobs service rest call when the job is executed. 

```
[kogito@sonataflow-platform-data-index-service-7c996f6579-cfw96 ~]$ curl -H "Content-Type: application/json" -H "Accept: application/json" -X POST --data '{"query" : "{ ProcessInstances { id, processId, state, endpoint, serviceUrl, start, end } }"  }' http://sonataflow-platform-data-index-service.usecase3/graphql
{"data":{"ProcessInstances":[{"id":"fe275052-f69c-470f-a2c5-6ed14ee7e16c","processId":"callbackstatetimeouts","state":"COMPLETED","endpoint":"http://callbackstatetimeouts.usecase3/callbackstatetimeouts","serviceUrl":"http://callbackstatetimeouts.usecase3","start":"2024-04-19T19:35:47.264Z","end":"2024-04-19T19:36:17.301Z"}]}
```

```
[kogito@sonataflow-platform-data-index-service-7c996f6579-cfw96 ~]$ curl -H "Content-Type: application/json" -H "Accept: application/json" -X POST --data '{"query" : "{ Jobs { id, processId, processInstanceId, status, expirationTime, retries, endpoint, callbackEndpoint } }"  }' http://sonataflow-platform-data-index-service/graphql
{"data":{"Jobs":[{"id":"4705f150-935f-4ba4-8ab7-d048d1a93f72","processId":"callbackstatetimeouts","processInstanceId":"fe275052-f69c-470f-a2c5-6ed14ee7e16c","status":"EXECUTED","expirationTime":null,"retries":0,"endpoint":"http://sonataflow-platform-jobs-service.usecase3/jobs","callbackEndpoint":"http://callbackstatetimeouts.usecase3:80/management/jobs/callbackstatetimeouts/instances/fe275052-f69c-470f-a2c5-6ed14ee7e16c/timers/-1"}]}}[kogito@sonataflow-platform-data-index-service-7c996f6579-cfw96 ~]$
```

wmedvede:
> @wmedvede Actually we should delete the workflow and recreate it. But we can explore this in a new issue.

I mean, a did a new deployment from scratch by removing the WF and deploying again with kubernetes deployment.

wmedvede:
Finally, it was possible to send a POST on a scaled to zero workflow and see a pod starting to attend the request and the WF instance was created.

wmedvede (approved):
I think it's working well, the only missing part is to fine tune the permissions as commented.

wmedvede:
btw, I have tested this in minikube, after last commit from @ricardozanini  **[Fix objects permissions and cluster roles]**

domhanak (approved):

wmedvede (commented):
@ricardozanini something is not good here.

If we deploy a workflow with persistence and knative deployment mode, the corresponding Service and Deployment is not created.


And we get always 

wmedvede@wmedvede techpreview2 (persistence-managed)$ kubectl get workflow -n usecase3-persistence
NAME                    PROFILE   VERSION   URL   READY   REASON
callbackstatetimeouts             0.0.1           False   DeploymentIsUnavailable


However, the same WF, with no persistence and  deploymentMode knative worked fine.

wmedvede:
@ricardozanini OK, I can see what the problem above is:

The workflow I'm using to test the persistence + deploymentMode : knative is like this the one below, see that it uses **initiContainers**, which is perfectly fine according with the SonataFlow spec. And this works fine for kubernetes deployments.

However, knative deployments **don't support initContainers**, and in the sonataflow-controller logs we can see this message:

E0508 11:02:32.582067       1 controller.go:324] sonataflow-manager "msg"="Reconciler error" "error"="admission webhook \"validation.webhook.serving.knative.dev\" denied the request: **validation failed: must not set the field(s): spec.template.spec.initContainers\npod spec support for init-containers is off, but found 1 init containers**: " "SonataFlow"={"name":"callbackstatetimeouts","namespace":"usecase3-persistence"} "controller"="sonataflow" "controllerGroup"="sonataflow.org" "controllerKind"="SonataFlow" "name"="callbackstatetimeouts" "namespace"="usecase3-persistence" "reconcileID"="808f0b61-f24c-428c-acbb-c989bc750554"

**I think that the SonataflowController should catch-up this situation and produce a validation error, and the documentation should be updated accordingly too.**

@ricardozanini feel free to add the validation on this PR or a followup one.

**Workflow used for testing with persistence:**

```
apiVersion: sonataflow.org/v1alpha08
kind: SonataFlow
metadata:
  name: callbackstatetimeouts
  annotations:
    sonataflow.org/description: Callback State Timeouts Example k8s
    sonataflow.org/version: 0.0.1
spec:
  persistence:
    postgresql:
      secretRef:
        name: postgres-secrets
        userKey: POSTGRES_USER
        passwordKey: POSTGRES_PASSWORD
      serviceRef:
        name: postgres
        databaseName: sonataflow
        databaseSchema: public
  podTemplate:
    deploymentModel: knative
    initContainers:
      - name: init-postgres
        image: registry.access.redhat.com/ubi9/ubi-minimal:latest
        imagePullPolicy: IfNotPresent
        command: [ 'sh', '-c', 'until (echo 1 > /dev/tcp/postgres.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local/5432) >/dev/null 2>&1; do echo "Waiting for postgres server"; sleep 3; done;' ]
    container:
      env:
        # by now, when the WF persistence is initialized we don't set migration to true.
        # we can re-evaluate to default to true.
        - name: QUARKUS_FLYWAY_MIGRATE_AT_START
          value: 'false'
  flow: bla bla bla

```

wmedvede (approved):
I have re-approved since after investigation, the suggested initiContaines validation is desirable but not a blocker.

ricardozanini:
@wmedvede noted; I'll update the docs to reflect the `initContainer` support for Knative. I'll merge this one for now and we can add this as an effort to https://github.com/apache/incubator-kie-kogito-serverless-operator/pull/239

PR #479, opened 2024-06-06
added isValidProfile to ProfileType

treblereel:
https://github.com/apache/incubator-kie-issues/issues/1302

added IsValidProfile() to ProfileType

Adding an IsValidProfile() to ProfileType in the serverless-operator would be very handy.


treblereel:
^^ @ricardozanini

ricardozanini (approved):

PR #482, opened 2024-06-06
Set recommended kubernetes labels on services and workflows

rgolangh:
## Motivation
There's a need to select efficiently workflows and their respective
services. Having the common kubernetes labels allow a single selector:

```yaml
   podSelector:
     matchExpressions:
       - { key: app.kubernetes.io/component, operator: In, values:
         ["data-index-service", "jobs-service", "serverless-workflow"] }
```

Modification:
Make the v1.Deployment for services and the deployment or knative
services to contain at common labels

Result:
A workflow deployment or knative serving labels:

```yaml
    app.kubernetes.io/name: ${workflow name}
    app.kubernetes.io/component: serverless-workflow
    app.kubernetes.io/part-of: ${platform url set by status}
    app.kubernetes.io/managed-by: sonataflow-operator
```

Data index or Jobs services Deployment.v1 labels:

```yaml
    app.kubernetes.io/name: ${service name}
    app.kubernetes.io/component: data-index-service|jobs-service
    app.kubernetes.io/part-of: ${platform name}
    app.kubernetes.io/managed-by: sonataflow-operator
```

Reference: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

Fixes: #483
Signed-off-by: Roy Golan <rgolan@redhat.com>


ricardozanini:
Thanks @rgolangh!

> A future improvement will be to pass the platform name to the workflow
so app.kubernetes.io/part-of will have the platform name so we can
select all component for a platform.

Can't you do this in this PR?

controllers/platform/k8s.go
```diff
@@ -233,8 +233,12 @@ func createOrUpdateService(ctx context.Context, client client.Client, platform *
...
 func getLabels(platform *operatorapi.SonataFlowPlatform, psh services.PlatformServiceHandler) (map[string]string, map[string]string) {
 	lbl := map[string]string{
-		workflowproj.LabelApp:     platform.Name,
-		workflowproj.LabelService: psh.GetServiceName(),
+		workflowproj.LabelApp:          platform.Name,
+		workflowproj.LabelService:      psh.GetServiceName(),
+		"app.kubernetes.io/name":       psh.GetServiceName(),
+		"app.kubernetes.io/component":  psh.GetContainerName(),
+		"app.kubernetes.io/part-of":    platform.Name,
+		"app.kubernetes.io/managed-by": "sonataflow-operator",
```
ricardozanini, line 241: Please add these to https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/api/metadata/annotations.go#L22

workflowproj/operator.go
```diff
@@ -85,8 +85,12 @@ func GetManagedPropertiesFileName(workflow *operatorapi.SonataFlow) string {
 // GetDefaultLabels gets the default labels based on the given workflow.
 func GetDefaultLabels(workflow *operatorapi.SonataFlow) map[string]string {
 	return map[string]string{
-		LabelApp:      workflow.Name,
-		LabelWorkflow: workflow.Name,
+		LabelApp:                       workflow.Name,
+		LabelWorkflow:                  workflow.Name,
+		"app.kubernetes.io/name":       workflow.Name,
+		"app.kubernetes.io/component":  "serverless-workflow",
+		"app.kubernetes.io/part-of":    workflow.Status.Services.DataIndexRef.Url,
```
ricardozanini, line 92: How hard it's to get the platform name here?

rgolangh:
For most of the code path it is easy except the various config maps
factories in workflowproj/operator.go
I can ignore the platform for those as a compromise

wmedvede (changes requested):
LGTM, just the same comment about te part-of

rgolangh:
> LGTM, just the same comment about te part-of

i would have to get the active platform set in the workflow reconciler and then pass it on to the rest of the state ensurers
some of them already do it i think
wdyt?

ricardozanini:
@rgolangh 
> For most of the code path it is easy except the various config maps
factories in workflowproj/operator.go

The `workflowproj` won't care about the platform AFAIK, so what we can do is to add a `Platform` argument to the constructors so a client can pass the platform (e.g. kn cli) when generating the manifests.

rgolangh:
I set it to wip as it became bit messy. Please give quick feedback if you can on the idea of that refactoring. 
I think I want to leave only a single ensurer that has the platform . same for the object creator. that should remove some uneeded code 
I had a place or two that are used just to generate the workflowproj so not in real deployment time. like @ricardozanini mentioned

ricardozanini:
@rgolangh this PR should ease your work on this PR:

- https://github.com/apache/incubator-kie-kogito-serverless-operator/pull/485

controllers/profiles/common/ensurer.go
```diff
@@ -35,7 +35,7 @@ var _ ObjectEnsurer = &noopObjectEnsurer{}
 var _ ObjectsEnsurer = &defaultObjectsEnsurer{}
 
 type ObjectEnsurer interface {
-	Ensure(ctx context.Context, workflow *operatorapi.SonataFlow, visitors ...MutateVisitor) (client.Object, controllerutil.OperationResult, error)
+	Ensure(ctx context.Context, workflow *operatorapi.SonataFlow, pl *operatorapi.SonataFlowPlatform, visitors ...MutateVisitor) (client.Object, controllerutil.OperationResult, error)
```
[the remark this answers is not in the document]
rgolangh: yes this was just an unclean attempt to see if this is useful. I think we can use the ensurer with the platform in all places now and remove the one without.

ricardozanini:
@rgolangh can you rebase and implement based on my PR?

rgolangh:
Thanks @zanini I rebased on your changes and rolled back most of the previous changes

ricardozanini:
@rgolangh can you please check the unit tests? Also, before sending the PR do:

```shell
make generate-all
make vet fmt
```

I need to update the CONTRIB guide.

ricardozanini:
@rgolangh can you check the tests?

rgolangh:
oh I now see that `make test` only tests the api/ folder and not the rest.
So I saw everything working locally and overlooked the PR

rgolangh:
> > oh I now see that `make test` only tests the api/ folder and not the rest.
> 
> wdym? `make test` runs everything, including the `api` module.

from the Makefile
```
test: manifests generate envtest vet fmt test-api ## Run tests.
...
test-api:
	cd api && make test

```

ricardozanini:
Yes, it runs everything.

rgolangh:
> Yes, it runs everything.


it stays on the api folder after the cd I think

controllers/platform/k8s.go
```diff
@@ -233,8 +233,12 @@ func createOrUpdateService(ctx context.Context, client client.Client, platform *
 
 func getLabels(platform *operatorapi.SonataFlowPlatform, psh services.PlatformServiceHandler) (map[string]string, map[string]string) {
 	lbl := map[string]string{
-		workflowproj.LabelApp:     platform.Name,
-		workflowproj.LabelService: psh.GetServiceName(),
+		workflowproj.LabelApp:          platform.Name,
```
ricardozanini, line 236: I think we can use the new k8s label in place of this. Can you remove it?

rgolangh: I think there are varioius selector on "app"
https://github.com/search?q=repo%3Aapache%2Fincubator-kie-kogito-serverless-operator+%5C%22app%5C%22&type=code

rgolangh: if this look okay I'll remove it

controllers/profiles/common/object_creators.go
```diff
@@ -76,6 +76,7 @@ const (
 // It serves as a basis for a basic Quarkus Java application, expected to listen on http 8080.
 func DeploymentCreator(workflow *operatorapi.SonataFlow, plf *operatorapi.SonataFlowPlatform) (client.Object, error) {
 	lbl := workflowproj.GetMergedLabels(workflow)
+	delete(lbl, workflowproj.LabelK8SPartOF)
```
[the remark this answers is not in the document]
rgolangh: I'll do that and in addition I think it is the best to also use the platform object we have there and not use the platform from the status of the workflow because it is not there - hence the errors in the controller log while trying to update the deployment selector
wdyt?

rgolangh:
Repeating a discussion we had on slack:
Since the selector labels are immutable we would constantly fail to reconcile existing deployments.
This means that existing deployments would fall out of the network policies based on the labels set here.

ricardozanini:
ok, just to add then we will remove from reconciliation the `labels` from the selector. Labels from other objects won't be replaced by the object management but MERGED instead. This way we avoid this kind of problem when migrating from different versions.

rgolangh:
> ok, just to add then we will remove from reconciliation the `labels` from the selector. Labels from other objects won't be replaced by the object management but MERGED instead. This way we avoid this kind of problem when migrating from different versions.

Ack this is probably the way to go. I'll make sure to adapt the code the add to the map and not replace.

rgolangh:
e2e test fixed and are green now.

one last note: installing sonata using an helm for example, would take over this label on all installed artifacts:
```
app.kubernetes.io/part-of: Helm
```
which ruins the nice query we thought about to get all the artifacts of a platform.

Fortunately according to helm  documentation this label usage is optional and not mandatory. So I suggest we change our helm installation to avoid this label completely. However the most important labels for now are the `app.kunernetes.io/component` with the values of `serverless-workflow, data-index-service, jobs-service`

rgolangh:
@wmedvede can you have a look?

wmedvede:
Hi @rgolangh , great work, many thanks.

Here goes some nitpicks I could find doing piece of testing. 

**Pod for a WF deployed with preview profile**

Part-of-preview-Screenshot from 2024-07-19 14-49-09

**Pod for a WF deployed with the dev profile**

Part-of-devmode-Screenshot from 2024-07-19 14-49-31


^ I couldn't check the gitops profile, please verify.


**Config maps for DI and JS**

Config-map-for-sevices-Screenshot from 2024-07-19 14-55-05

**This is the config map for a workflow**

Config-maps-workflow-1-Screenshot from 2024-07-19 14-56-54

**Managed properties config map**

Config-maps-workflow2-Screenshot from 2024-07-19 15-00-07

**Workflow deployed with the dev profile, both config maps was created by the operator:**

Config-map-created-by-operator-Screenshot from 2024-07-19 15-08-01

**But the managed properties for same WF, no labels.**
Config-map-created-by-operator-Screenshot from 2024-07-19 15-10-25


**When we create a WF with the dev profile, a route is created for it**

Route-dev-mode-Screenshot from 2024-07-19 15-04-23

**Labels for builds**
When a WF is created in Openshift with the preview mode, the following objects are created.

A-build-config-Screenshot from 2024-07-19 15-26-36


A-potential-build-Screenshot from 2024-07-19 15-28-27

A-potential-image-stream-Screenshot from 2024-07-19 15-30-45


A-pod-build-Screenshot from 2024-07-19 15-36-18

wmedvede (changes requested):
Code looks good, many thanks.
I have left some testing nitpicks.

rgolangh:
to summarize the artifacts the operator creates directly or indirectly with no lables:
- the build pod
- route in dev prof without part-of 
- managedProps without labels
- pod in dev prof without part-of

wmedvede:
> to summarize the artifacts the operator creates directly or indirectly with no lables:
> 
> * the build pod
> * route in dev prof without part-of
> * managedProps
> * pod in dev prof without part-of

Yes.
Then, one thing that will look weird is to have the managed props, created by the operator, labeled, and the "standard" props (user provided) not labeled. Also, regarding the "standard" props, we have two situations.

1) the user never provided the "standard" props. In that case, the operator still creates an empty one, this is good for consistency. In this case, the operator can label them.

2) the user has provided the "standard" props. Here, the operator should label them?
From the user perspective might look weird to see the config map he provided to gain some labels automatically after some time :thinking: , but on the other hand, for consistence we must label them.



, the thing that will look weird is to have managed props for a given wo

rgolangh:
I think it would be excellent to label the props the user provided, because that really binds a workflow and its dependencies nicely together.

wmedvede:
@ricardozanini 

I'm approving and merging this with the gaps stated in the screenshots above to not block another feature that is being developed by  @rgolangh on top of this labels.

@rgolangh Would you mind open an issue to keep track of the gaps that are going be left so this can be tackled at some point.

wmedvede (approved):

rgolangh:
@wmedvede created https://github.com/apache/incubator-kie-kogito-serverless-operator/issues/509

rgolangh:
#508 Add tests to cover workflowproj/operator.go

PR #535, opened 2024-09-20
 kie-kogito-serverless-operator-405: Add external built image integri…

treblereel:
issue: https://github.com/apache/incubator-kie-kogito-serverless-operator/issues/405

ps: tests will be added in the next commit to this pr


treblereel:
[sonataflow_image.yaml.zip](https://github.com/user-attachments/files/17069620/sonataflow_image.yaml.zip)

this yaml can be used to test the PR. It contained image declaration with sonataflow project, published to DH

internal/controller/validation/ImageValidator.go
```diff
@@ -0,0 +1,123 @@
...
+
+		if header.Typeflag == '0' && header.Name == workflow {
+			return tarReader, nil
+		}
+	}
+
+	return nil, nil
+}
+
+func jsonFromDockerImage(reader io.Reader) (operatorapi.Flow, error) {
```
ricardozanini, line 116:
```suggestion
func workflowSpecFromDockerImage(reader io.Reader) (operatorapi.Flow, error) {
```

I believe this name reflects better what it does. WDYT?

internal/controller/validation/ImageValidator.go
```diff
@@ -0,0 +1,123 @@
...
+		if reader, err := readLayer(layers[i], workflow); err == nil && reader != nil {
+			return reader, nil
+		} else if err != nil {
+			return nil, err
+		}
+	}
+	return nil, fmt.Errorf("file not found %s in docker image", workflow)
+}
+
+func readLayer(layer v1.Layer, workflow string) (*tar.Reader, error) {
```
ricardozanini, line 91:
```suggestion
func findWorkflowSpecLayer(layer v1.Layer, workflow string) (*tar.Reader, error) {
```

internal/controller/validation/ImageValidator.go
```diff
@@ -0,0 +1,123 @@
...
+
+	workflowDockerImage, err := jsonFromDockerImage(reader)
+	if err != nil {
+		return false, err
+	}
+
+	return cmp.Equal(workflowDockerImage, sonataflow.Spec.Flow), nil
+}
+
+func readLayers(image v1.Image, workflow string) (*tar.Reader, error) {
```
ricardozanini, line 75:
```suggestion
func readWorkflowSpecLayer(image v1.Image, workflow string) (*tar.Reader, error) {
```

The name of this `func` resembles an array return. WDYT about the suggested name?

internal/controller/validation/ImageValidator.go
```diff
@@ -0,0 +1,123 @@
...
+			return err
+		}
+		if !equals {
+			return fmt.Errorf("Workflow, defined in the image %s doesn't match deployment workflow", sonataflow.Spec.PodTemplate.Container.Image)
+		}
+	}
+	return nil
+}
+
+func validateImage(sonataflow *operatorapi.SonataFlow, image string) (bool, error) {
```
ricardozanini, line 51: Can you move all these funcs to the `ImageValidator` struct? We can keep them private and avoid mixing names when new validators come in.

ricardozanini (commented):
It seems pretty nice!

ricardozanini:
The e2e tests have been fixed, please rebase your branch 🙏

treblereel:
> The e2e tests have been fixed, please rebase your branch 🙏
Thanks

Unfortunately, I have to fix `test/testdata/sonataflow.org_v1alpha08_sonataflow-simpleops.yaml` first, because of       `image: quay.io/kiegroup/sonataflow-minimal-example:latest`

ricardozanini:
@treblereel I'm currently working to fix this one.

treblereel:
> sonataflow-minimal-example

The problem here is sonataflow-minimal-example has no workflow.sw.json in it, so I have to build and publish it to temp repo before tests ...

treblereel:
Currently, to verify that the image corresponds to the deployment, it is checked that `/deployments/app/workflow.sw.json` in the image matches the workflow in the YAML.

Comments:

- This approach aligns with how `kn-workflow-plugin` generates new projects.
- If this approach is acceptable, `kn-workflow-plugin` needs to be updated (update Dockerfile.*).
- Other naming options are possible, such as `metadata.name`.
- It's breaking change, if it is merged, all deployments with image declaration will fail

An additional step has been added that replaces `kind-registry` in `spec.podTemplate.container.image` with the IP address of kind-registry in `*operatorapi.SonataFlow` during the deployment phase, because the container builder cannot resolve the image URI by hostname in it's `kind-registry`. This issue does not occur with an external Docker registry.

.github/workflows/e2e.yml
```diff
@@ -76,6 +76,10 @@ jobs:
           make container-build BUILDER=docker IMG=${{ env.OPERATOR_IMAGE_NAME }}
           kind load docker-image ${{ env.OPERATOR_IMAGE_NAME }}
 
+      - name: Build and Install SonataFlow worfklow test images
+        run: |
+          make workflow_test_image_build-and-push
```
ricardozanini, line 81: This is unnecessary. The e2e `BeforeSuite` uses the operator to deploy three workflows to build images within the cluster. At the end of the build, the suite will fetch the images from the `SonataFlowBuild` `.status.image`. 

During the e2e, we replace the `e2e/testdata` YAML `.spec.podTemplate.container.image` attribute. So that when the operator deploys the `SonataFlow` custom resource the image is already there in the registry.

If you were to add a new e2e make sure to use the [kubectlPatchSonataFlowImageAndRollout](https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/test/e2e/helpers.go#L97) to replace with the right image tag:

```go
type E2EWorkflows struct {
	// CallBackPersistence is the prebuilt workflow to use on every use case that requires persistence
	CallBackPersistence E2EWorkflowEntry
	// CallBack is the prebuilt workflow to use on every use case that requires events or jobs services integration
	CallBack E2EWorkflowEntry
	// Greetings is the general usage image to use on any use case that requires a SonataFlow deployment
	Greetings E2EWorkflowEntry
}
```

internal/controller/validation/image_url_sanitizer.go
```diff
@@ -0,0 +1,69 @@
...
+	operatorapi "github.com/apache/incubator-kie-kogito-serverless-operator/api/v1alpha08"
+	"github.com/google/go-containerregistry/pkg/name"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type imageUrlSanitizer struct{}
+
+func (v *imageUrlSanitizer) Validate(ctx context.Context, client client.Client, sonataflow *operatorapi.SonataFlow, req ctrl.Request) error {
+	isInKindRegistry, kindRegistryUrl, err := imageStoredInKindRegistry(ctx, sonataflow.Spec.PodTemplate.Container.Image)
```
ricardozanini, line 31: I don't think this is necessary; the docker API should be enough since Kind implements OCI.

internal/controller/validation/image_validator.go
```diff
@@ -0,0 +1,165 @@
...
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type imageValidator struct{}
+
+const workflowName = "deployments/app/workflow.sw.json"
```
ricardozanini, line 37:
```suggestion
const workflowPath = "deployments/app/workflow.sw.json"
```

ricardozanini: The workflow might be YAML, too. And we can't fix the file name. IIRC the file name is the same as `.metadata.name`. Investigation needed.

test/testdata/sonataflow.org_v1alpha08_sonataflow-simpleops.yaml
```diff
@@ -26,7 +26,7 @@ spec:
     container:
       # TODO use docker.io/apache/incubator-kie-sonataflow-minimal-example:latest
       # when we complete https://github.com/apache/incubator-kie-kogito-serverless-operator/issues/504
-      image: quay.io/kiegroup/sonataflow-minimal-example:latest
+      image: kind-registry:5000/testimage/sonataflow-minimal-example:0.1
```
ricardozanini, line 29: We should replace this in runtime based on the result of the `SonataFlowBuild` status.

treblereel: yeap, I'll remove it all, since we are not going to push images directly to the registry

internal/controller/validation/image_validator.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	return nil
+}
+
+func NewImageValidator() Validator {
+	return &imageValidator{}
+}
+
+func validateImage(ctx context.Context, sonataflow *operatorapi.SonataFlow) (bool, error) {
+	isInKindRegistry, _, err := imageStoredInKindRegistry(ctx, sonataflow.Spec.PodTemplate.Container.Image)
```
ricardozanini, line 59: Why do we need this? How's the kind of registry different? AFAIK, the image has the same tag as any other registry.

ricardozanini: Anyhow, it's nice that we support this difference.

internal/controller/validation/common.go
```diff
@@ -0,0 +1,117 @@
...
+	"strings"
+
+	"github.com/apache/incubator-kie-kogito-serverless-operator/internal/controller/platform"
+	"github.com/apache/incubator-kie-kogito-serverless-operator/utils"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var kindRegistryName = func(ctx context.Context) (string, error) {
```
ricardozanini, line 31: Why does it need to be a var?

ricardozanini (commented):
Also please remove the `container-builder/bin` folder. Not sure why they are there. 🤔

ricardozanini:
@treblereel I think we are closer to merging this PR. Can you please also provide the docs on `incubator-kie-kogito-docs`? I believe this feature deserves a new page under `Cloud`. 

@wmedvede @jakubschwan , do you mind taking a look?

internal/controller/validation/common.go
treblereel: @ricardozanini for mock purpose, to resolve kind-registry in the tests

internal/controller/validation/image_validator.go
treblereel: @ricardozanini I use `go-containerregistry` to fetch images from the docker registry. Kind-registry is insecure, so I have to force `go-containerregistry` to skip verifications. `imageStoredInKindRegistry()` is used to solce if the image is on kind-registry, or not.

ricardozanini (commented):
@treblereel controller-gen is still there. Please also add an entry to .gitignore `bin`.

In the docs, we must add details of this verification and instruct users how to build or customize a dockerfile to build a valid image.

ricardozanini:
Waiting for QE verification @jakubschwan and @wmedvede's review.

@treblereel the docs can be pushed later.

jakubschwan (commented):
Changes in PR looks good to me

treblereel:
> Waiting for QE verification @jakubschwan and @wmedvede's review.
> 
> @treblereel the docs can be pushed later.

https://github.com/apache/incubator-kie-kogito-docs/pull/673

wmedvede (changes requested):

wmedvede:
Hi @treblereel , I have left some minor comments/nitpicks.

Before merging I thing guys @jakubschwan @treblereel we must be sure this was tested in:
1) minikube
2) openshift

and also using images docker.io and quay.io.

Many thanks!

internal/controller/validation/validator.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2024 Apache Software Foundation (ASF)
```
wmedvede, line 1: Same comment regarding license

internal/controller/validation/image_validator.go
```diff
@@ -0,0 +1,169 @@
+// Copyright 2024 Apache Software Foundation (ASF)
```
wmedvede, line 1: Same comment regarding license

internal/controller/validation/common_test.go
```diff
@@ -0,0 +1,118 @@
+// Copyright 2024 Apache Software Foundation (ASF)
```
wmedvede, line 1: Same comment regarding license

wmedvede (changes requested):

internal/controller/validation/common.go
```diff
@@ -0,0 +1,117 @@
+// Copyright 2024 Apache Software Foundation (ASF)
```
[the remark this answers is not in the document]
treblereel: @wmedvede I ran `make addheaders`, which add //, anyway, it's fixed atm

internal/controller/sonataflow_controller.go
```diff
@@ -98,6 +100,11 @@ func (r *SonataFlowReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, err
 	}
 
+	if err := r.Validate(ctx, workflow, req); err != nil {
+		klog.V(log.E).ErrorS(err, "Failed to validate SonataFlow")
+		return reconcile.Result{}, nil
```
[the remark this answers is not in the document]
treblereel: @wmedvede @ricardozanini 

This is quite a tricky situation: if we return the error, the deployment will be retried until the limit is reached. A common pattern to solve this problem is to return from Reconcile and log the error to the console.

internal/controller/validation/image_validator.go
```diff
@@ -0,0 +1,169 @@
...
+	isInKindRegistry, _, err := imageStoredInKindRegistry(ctx, sonataflow.Spec.PodTemplate.Container.Image)
+	if err != nil {
+		return false, err
+	}
+
+	var ref v1.Image
+	if isInKindRegistry {
+		ref, err = kindRegistryImage(sonataflow)
+	} else {
+		ref, err = remoteImage(sonataflow)
```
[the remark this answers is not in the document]
treblereel: @wmedvede @ricardozanini 

The whole point of the `isInKindRegistry` check is to determine whether the image is in the `kind-registry` or not. As far as I understand, this only makes sense for tests and only in the case where the 'data.localRegistryHosting.v1.hostFromClusterNetwork' flag is set in the `ConfigMap`. Kind-registry is insecure. Overall, I don't like the idea of adding checks to the production code just to simplify testing (I'd prefer to remove this check), but the strategy for fetching an image from a secure and insecure registry is different

One option would be to add certificates (self-signed ones are not suitable) to make the `kind-registry` secure

This does not interfere with `minikube` because we don't use `kind-registry` in it

For testing, I use this image from Docker: `docker.io/treblereel/sonataflow-workflow-demo:latest`. I don't think there should be any issues with quay.io, but I will check further

internal/controller/validation/image_validator.go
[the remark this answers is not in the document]
treblereel: @wmedvede As I can see, we use kind-registry and `sonataflow.Spec.PodTemplate.Container.Image` only in the tests, where `data.localRegistryHosting.v1.hostFromClusterNetwork` set. In case we want to use the same scenario with the minikube, it's better to add to `install-minikube.sh` registry and ConfigMap from the `create-kind-cluster-with-registry.sh`.

internal/controller/validation/image_validator.go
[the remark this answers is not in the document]
treblereel: @wmedvede quay.io works 
[sonataflow_image_quay.yaml.zip](https://github.com/user-attachments/files/17399989/sonataflow_image_quay.yaml.zip)

PR #539, opened 2024-09-23
[issue-464]Create a Prometheus ServiceMonitor object that can capture/collect metrics from deployed SonataFlow instances

jianrongzhang89:
This PR implements https://github.com/apache/incubator-kie-kogito-serverless-operator/issues/464.

The operator an track in the cluster if Prometheus Operator is installed. If this condition is true and the SonataFlowPlatform is marked with .spec.monitoringEnabled: true, then each workflow instance can have a [ServiceMonitor](https://prometheus-operator.dev/docs/operator/design/#servicemonitor) object bound to it.

**Checklist**
- [X ] Add or Modify a unit test for your change


jianrongzhang89:
Not needed anymore.

PR #551, opened 2024-10-14
kie-kogito-serverless-operator-548: Make the operator configure the kogito events grouping by default in preview and gitops workflow deployments

wmedvede:

**Checklist**

- [X] Add or Modify a unit test for your change
- [X] Have you verified that tall the tests are passing?

<details>
<summary>
</summary>

</details>

### DIFF
--- a/bundle/manifests/sonataflow-operator-controllers-config_v1_configmap.yaml
+++ b/bundle/manifests/sonataflow-operator-controllers-config_v1_configmap.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 data:
-  controllers_cfg.yaml: |
+  controllers_cfg.yaml: |-
     # Licensed to the Apache Software Foundation (ASF) under one
     # or more contributor license agreements.  See the NOTICE file
     # distributed with this work for additional information
@@ -58,6 +58,9 @@ data:
       - groupId: org.kie
         artifactId: kie-addons-quarkus-persistence-jdbc
         version: 999-20240912-SNAPSHOT
+    # If true, the workflow deployments will be configured to send accumulated workflow status change events to the Data
+    # Index Service reducing the number of produced events. Set to false to send individual events.
+    kogitoEventsGrouping: true
 kind: ConfigMap
 metadata:
   name: sonataflow-operator-controllers-config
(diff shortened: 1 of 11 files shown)


internal/controller/platform/services/services.go
```diff
@@ -603,6 +603,7 @@ func (d *DataIndexHandler) GenerateKnativeResources(platform *operatorapi.Sonata
 		d.newTrigger(lbl, brokerName, namespace, serviceName, "process-state", "ProcessInstanceStateDataEvent", constants.KogitoProcessInstancesEventsPath, platform),
 		d.newTrigger(lbl, brokerName, namespace, serviceName, "process-variable", "ProcessInstanceVariableDataEvent", constants.KogitoProcessInstancesEventsPath, platform),
 		d.newTrigger(lbl, brokerName, namespace, serviceName, "process-definition", "ProcessDefinitionEvent", constants.KogitoProcessDefinitionsEventsPath, platform),
+		d.newTrigger(lbl, brokerName, namespace, serviceName, "process-instance-multiple", "MultipleProcessInstanceDataEvent", constants.KogitoProcessInstancesMultiEventsPath, platform),
```
ricardozanini, line 606: What happens if I toggle to `false` and then to `true` again? Are we recreating/deleting the trigger instance?

wmedvede: This basically adds a new trigger, with the same pattern for the other existing ones.

@jianrongzhang89 do you know what happens in the situation introduced by @ricardozanini ?

jianrongzhang89: @ricardozanini current implementation does not delete the existing triggers if it the flag is set from true to false (and they will get deleted if the CR is deleted) and when it is to true again - the new trigger gets added.

ricardozanini: So we will have dangling triggers in case we turn it to false?

jianrongzhang89: Yes.

bundle/manifests/sonataflow-operator-controllers-config_v1_configmap.yaml
```diff
@@ -58,6 +58,9 @@ data:
       - groupId: org.kie
         artifactId: kie-addons-quarkus-persistence-jdbc
         version: 999-20240912-SNAPSHOT
+    # If true, the workflow deployments will be configured to send accumulated workflow status change events to the Data
+    # Index Service reducing the number of produced events. Set to false to send individual events.
+    kogitoEventsGrouping: true
```
ricardozanini, line 63: Why are we setting this on the controller and not the on platform level? The relationship between the platform and data index is 1:1, so this information should be set there in this case.

wmedvede: @ricardozanini I don't want to contaminate the SPF definition with that value, so, in the sonataflowplatform CRD definition we don't add this value, like we did with the dbMigrator parameter.

Instead we have this.

1. At cluster level, people can enable/disable the kogitoEventsGrouping. By default I configured the value true. Since after discussions in our meetings and with @fjtirado it looks like the best suited configuration. So by default, based on the kogitoEventsGrouping, we'll add the corresponding value to the corresponding application.properties.
2. BUT, if people wants to overwrite this value we set by default, for whatever reason, they can go to the corresponding WF, DI, or JS configuraion and set the equivalent env variable, that will overwrite the value the opeator set in the application.properties.

So, with 1 and 2, we have a simple mechanims to configure the kogitoEvensGrouping by default, and, people can always overwrite in particular situations. Which, as it was investigated, is not the recommended approach.

ricardozanini: It would be okay if users could overwrite this value via ENV or properties. 👍

ricardozanini (approved):

PR #553, opened 2024-10-16
[issue-547] Knative Broker is required even if is not defined in SonataFlow or SonataFlowPlatform

### LINKED ISSUE #547: Knative Broker is required even if is not defined in `SonataFlow` or `SonataFlowPlatform`
### Describe the bug

After #368, the Knative Broker is required to wire together Jobs Service, Data Index, and workflows. Before, we used the direct wire configuration, with all services sending CEs directly to the HTTP listeners. This setup is far from ideal for a production use case but fair to use in a trial or development environment. 

This way, we restrict the context to which Knative must be present in order to run use cases with the supporting services.

### Summary

1. If the Knative Broker is not configured, the operator should fall to the old way of sending direct CEs to the services' endpoint
2. If the Knative Broker is indeed required either in the platform or workflows, we should fail if it's not present in the target namespace

This way, users can still deploy the services and workflows together despite the Knative installation.

Relates to:

- #541 
- #368 

### Actual behavior

The Operator requires Knative Broker to be present despite the workflow or platform configuration.

jianrongzhang89:
This PR implements the following when Knative is installed:
1) When there is no sink defined in a workflow, and there is no broker defined in the platform, the workflow reconciliation will continue instead of failing with errors. 
2) When a workflow does not have a broker defined in the workflow's spec.sources for the event type for events that the workflow consumes, and there is no broker configured in the platform, the workflow reconciliation will continue instead of failing with errors. In this case, no trigger will be created for that event type.

This is for backward compatibility with the latest operator release and usability. If will be up to the user to decide whether a broker should be configured for a workflow.

**Checklist**

- [ X] Add or Modify a unit test for your change
- [ X] Have you verified that tall the tests are passing?

Fix #547


ricardozanini (commented):
It seems pretty straightforward. Can you modify or add an e2e that validates it?

test/e2e/clusterplatform_test.go
```diff
@@ -230,7 +230,8 @@ var _ = Describe("Cluster Platform Use Cases :: ", Label("cluster"), Ordered, fu
 			Expect(err).NotTo(HaveOccurred())
 		},
 			Entry("without services configured", test.GetPathFromE2EDirectory("platform", "noservices"), metadata.GitOpsProfile.String(), ephemeral, false),
-			Entry("with services configured", test.GetPathFromE2EDirectory("platform", "services"), metadata.GitOpsProfile.String(), "ephemeral-with-workflow", true),
+			Entry("with services configured and platform broker", test.GetPathFromE2EDirectory("platform", "services"), metadata.GitOpsProfile.String(), "ephemeral-with-workflow", true),
+			Entry("with services configured and no broker", test.GetPathFromE2EDirectory("platform", "services"), metadata.GitOpsProfile.String(), "ephemeral-with-workflow-no-broker", true),
```
ricardozanini, line 234: Since this test does not have a `SonataFlowClusterPlatform`, I think the e2e label might be incorrect.

ricardozanini (approved):
Please disregard my comment. The new test uses a clustered platform.

ricardozanini:
@wmedvede @jakubschwan can you please verify it?

wmedvede (approved):

PR #558, opened 2024-10-23
[issue-521] Enhance error messages in SonataFlowPlatform CR status for knative integration

jianrongzhang89:
Implement https://github.com/apache/incubator-kie-kogito-serverless-operator/issues/521.
Note that we have decided to use K8s events instead of status conditions to notify the user that the broker is not available.

To facilitate troubleshooting of configuration issues.

**Checklist**

- [ X] Add or Modify a unit test for your change


ricardozanini (approved):

internal/controller/platform/k8s.go
```diff
@@ -359,18 +359,24 @@ func createOrUpdateKnativeResources(ctx context.Context, client client.Client, p
...
+				return nil, err
 			}
 			kSinkInjected, err := psh.CheckKSinkInjected()
 			if err != nil {
-				return err
+				return nil, err
 			}
 			if !kSinkInjected {
-				return fmt.Errorf("waiting for K_SINK injection for %s to complete", psh.GetServiceName())
+				msg := fmt.Sprintf("waiting for K_SINK injection for %s to complete", psh.GetServiceName())
```
wmedvede, line 369: This is good, we can see the service name in the event.

internal/controller/platform/services/services.go
```diff
@@ -582,18 +583,23 @@ func (d *DataIndexHandler) newTrigger(labels map[string]string, brokerName, name
...
 	namespace := broker.Ref.Namespace
 	if len(namespace) == 0 {
 		namespace = platform.Namespace
 	}
 	if err := knative.ValidateBroker(brokerName, namespace); err != nil {
-		return nil, err
+		event := &corev1.Event{
+			Type:    corev1.EventTypeWarning,
+			Reason:  WaitingKnativeEventing,
+			Message: err.Error(),
```
wmedvede, line 600: For this error message it would be nice if we can see the service name as part of the message.

I have forced the use case to happen, and in the events I can see:

5m47s       Warning   WaitingKnativeEventing   sonataflowplatform/sonataflow-platform                         broker sonataflow-broker in namespace case1-kn-eventing does not exist
13m         Warning   Failed                   sonataflowplatform/sonataflow-platform                         Failed to update SonataFlowPlaform: broker sonataflow-broker in namespace case1-kn-

So, it's difficult to distinguish for which service we have the missing broker.

If it's not too much work, it would include the we could include the name of the service as part of the message.
I think we can just grab the ValidatBroker message and concatenate "service" + d.GetServiceName()

and we can see:

sonataflow-broker in namespace case1-kn-eventing does not exist for **service: sonataflow-platform-data-index**

Considering that the configuration alternatives admits different Brokers for DI and JS, I think it'll be very helpful in situations where users needs to query the events to find the cause of the issue

wmedvede (approved):
@jianrongzhang89 @ricardozanini 
A minor non-blocking suggestion from my side guys if we have time.

internal/controller/platform/services/services.go
wmedvede: Same for the jobs-service knative addings

wmedvede: Many thanks!

PR #564, opened 2024-10-31
Fix #560 - Check if the .spec.flow has changed before building

### LINKED ISSUE #560: Add the ability to scale to 0 a workflow executing with the preview by not requiring a new build
### Description

Right now, while it's possible to scale to zero a workflow in preview profile, when the spec.podTemplate.repclias in the SonataFlow CRD is set to 0, prior to removing the pods, a new build is produced. No new build is needed in this case.

### Implementation ideas

_No response_

ricardozanini:
Fix #560 

**Checklist**

<details>
<summary>
</summary>

</details>

### DIFF
--- a/api/v1alpha08/sonataflow_types.go
+++ b/api/v1alpha08/sonataflow_types.go
@@ -200,6 +200,8 @@ type SonataFlowStatus struct {
 	// Triggers list of triggers created for the SonataFlow
 	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="triggers"
 	Triggers []SonataFlowTriggerRef `json:"triggers,omitempty"`
+	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="flowRevision"
+	FlowCRC uint32 `json:"flowCRC,omitempty"`
 }
 
 // SonataFlowTriggerRef defines a trigger created for the SonataFlow.
--- a/bundle/manifests/sonataflow.org_sonataflows.yaml
+++ b/bundle/manifests/sonataflow.org_sonataflows.yaml
@@ -10050,6 +10050,9 @@ spec:
               endpoint:
                 description: Endpoint is an externally accessible URL of the workflow
                 type: string
+              flowCRC:
+                format: int32
+                type: integer
               lastTimeRecoverAttempt:
                 format: date-time
                 type: string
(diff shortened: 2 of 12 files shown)


wmedvede (approved):

wmedvede (commented):
I have created a WF, I can see the build as usual.
I have changed the WF definition, and executed kubectl apply -f xxxx again.
I can't see a new build.

@ricardozanini looks like not working.

ricardozanini:
@wmedvede That's because we received the `.spec.flow` already updated in the k8s database. It would only be possible to determine if a workflow would change. I am not sure why the first version worked on my minikube.

Now, I've added a CRC field in the status, so we know if the flow has changed since we control the status update, not the user.

wmedvede (changes requested):
I have executed the following sequence:

1. I deployed a workflow, the build_1 is produced, the workflow pod is started, etc.
2. I did a breaking change in the workflow, and kubectl apply -f bla bla (I can see the new CRC generated)
3. A new build_2 is stated as expected
4. the new build_2 fails as expeced (see 2)
5. I fixed the workflow, and did kubectl apply -f bla bla
6. I can see the new CRC was generated, however no new build was started
7. I do more changes in the workflow, kubectl apply bla bla bla again
8. I can see the new CRC is generated

However, since the build_2, no new build is produced.
Not good.

ricardozanini:
@wmedvede, can you share the `SonataFlow` YAML for each change you made?

wmedvede:
> @wmedvede, can you share the `SonataFlow` YAML for each change you made?
> 
> I'm just wondering if this behavior happened before this PR, so the bug might be somewhere else in the `states_preview` algorithm.

I believe the issue us reproducible for whatever workflow, but you can use this use case:

https://github.com/flows-examples/techpreview2/blob/7103145b9c7917c31991e03607c52a0615348ec6/platforms/openshift-dataindex_and_jobservice_as_platform_service_and_global_persistence_postgresql/Readme.txt

Install the platform as pointed out, and then deploy the callback state timeouts.

https://github.com/flows-examples/techpreview2/blob/7103145b9c7917c31991e03607c52a0615348ec6/platforms/openshift-dataindex_and_jobservice_as_platform_service_and_global_persistence_postgresql/04-sonataflow_callbackstatetimeouts.sw.yaml

1) The first deploy will work
2) To force the second build, basically change:
```
  flow:
    start: PrintStartMessage
```

to the value

```
  flow:
    start: PrintStartMessage2
```

The second build will start, but will fail. Since the sate definition wasn't changed.

3) For the third build, keep:

```
  flow:
    start: PrintStartMessage2
```

and go to the workflows states, and change the name of the state also there:

```
    states:
      - name: PrintStartMessage2
        type: operation
        actions:
```

This time, a potential new build should be good.
However, it's never produced.

wmedvede (approved):

PR #565, opened 2024-11-04
[issue-562] Serverless workflow pod gets restarted repeatedly after native K_SINK injection

jianrongzhang89:
Fix issue https://github.com/apache/incubator-kie-kogito-serverless-operator/issues/562.

The reason why the ***00001-deployment*** pod gets restarted repeatedly is because after K_SINK injection, a new Knative revision has been created, and the workflow managed properties configmap has been updated to use the the K_SINK environment variable. As such the first deployment ****_00001 has become outdated and no longer aligned with the configmap and this causes the repeated pod restarting.

The fix is to clean up the outdated Knative revision(s) after the K_SINK injection is done (ie, sinkbinding is ready).

To prevent ***00001-deployment*** pod from getting restarted repeatedly.

**Checklist**

- [X ] Add or Modify a unit test for your change 

### DIFF
--- a/internal/controller/profiles/common/knative_eventing.go
+++ b/internal/controller/profiles/common/knative_eventing.go
@@ -20,9 +20,9 @@ package common
 import (
 	"context"
 
+	operatorapi "github.com/apache/incubator-kie-kogito-serverless-operator/api/v1alpha08"
 	"k8s.io/klog/v2"
 
-	operatorapi "github.com/apache/incubator-kie-kogito-serverless-operator/api/v1alpha08"
 	"github.com/apache/incubator-kie-kogito-serverless-operator/internal/controller/knative"
 	"github.com/apache/incubator-kie-kogito-serverless-operator/log"
 	"sigs.k8s.io/controller-runtime/pkg/client"
--- a/internal/controller/profiles/preview/states_preview.go
+++ b/internal/controller/profiles/preview/states_preview.go
@@ -22,21 +22,31 @@ package preview
 import (
 	"context"
 	"fmt"
+	"sort"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/klog/v2"
+	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/apache/incubator-kie-kogito-serverless-operator/api"
 	operatorapi "github.com/apache/incubator-kie-kogito-serverless-operator/api/v1alpha08"
 	"github.com/apache/incubator-kie-kogito-serverless-operator/internal/controller/builder"
+	"github.com/apache/incubator-kie-kogito-serverless-operator/internal/controller/knative"
 	"github.com/apache/incubator-kie-kogito-serverless-operator/internal/controller/platform"
 	"github.com/apache/incubator-kie-kogito-serverless-operator/internal/controller/profiles/common"
 	"github.com/apache/incubator-kie-kogito-serverless-operator/internal/controller/profiles/common/constants"
 	"github.com/apache/incubator-kie-kogito-serverless-operator/log"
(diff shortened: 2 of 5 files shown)


internal/controller/profiles/common/knative_eventing.go
```diff
@@ -79,3 +85,37 @@ func (k knativeObjectManager) Ensure(ctx context.Context, workflow *operatorapi.
...
+		LabelSelector: labels.SelectorFromSet(
+			map[string]string{
+				workflowproj.LabelWorkflow:          workflow.Name,
+				workflowproj.LabelWorkflowNamespace: workflow.Namespace,
+			},
+		),
+		Namespace: workflow.Namespace,
+	}
+	revisionList := &servingv1.RevisionList{}
+	if err := k.C.List(context.TODO(), revisionList, opts); err != nil {
```
ricardozanini, line 110:
```suggestion
	if err := k.C.List(ctx, revisionList, opts); err != nil {
```

internal/controller/profiles/common/knative_eventing.go
```diff
@@ -67,6 +70,9 @@ func (k knativeObjectManager) Ensure(ctx context.Context, workflow *operatorapi.
 			return objs, err
 		} else if sinkBinding != nil {
 			objs = append(objs, sinkBinding)
+			if err := k.cleanupOutdatedRevisions(ctx, workflow, sinkBinding); err != nil {
```
ricardozanini, line 73: We have a post-hook implementation that we can use to garbage collect it. WDYT? Also, `sinkBinding` is not used anywhere.

https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/internal/controller/profiles/profile.go#L74

So, in the reconciliation states, we can have a GC calling `cleanupOutdatedRevisions`.

jianrongzhang89: > We have a post-hook implementation that we can use to garbage collect it. WDYT? Also, `sinkBinding` is not used anywhere.
> 
> https://github.com/apache/incubator-kie-kogito-serverless-operator/blob/main/internal/controller/profiles/profile.go#L74
> 
> So, in the reconciliation states, we can have a GC calling `cleanupOutdatedRevisions`.

Sounds good. PR updated.

internal/controller/profiles/common/knative_eventing.go
jianrongzhang89: OK.

internal/controller/profiles/preview/states_preview.go
```diff
@@ -233,3 +237,44 @@ func (h *deployWithBuildWorkflowState) isWorkflowChanged(workflow *operatorapi.S
...
+	}
+	if !avail.Serving || !avail.Eventing {
+		return nil
+	}
+	injected, err := knative.CheckKSinkInjected(workflow.Name, workflow.Namespace)
+	if err != nil {
+		return err
+	}
+	if !injected {
+		return fmt.Errorf("waiting for Sinkbinding K_SINK injection to complete")
```
ricardozanini, line 257: Why are we returning an error here? Just to requeue?

jianrongzhang89: @ricardozanini yes this is to requeue.

internal/controller/profiles/preview/states_preview.go
```diff
@@ -233,3 +242,61 @@ func (h *deployWithBuildWorkflowState) isWorkflowChanged(workflow *operatorapi.S
...
+		LabelSelector: labels.SelectorFromSet(
+			map[string]string{
+				workflowproj.LabelWorkflow:          workflow.Name,
+				workflowproj.LabelWorkflowNamespace: workflow.Namespace,
+			},
+		),
+		Namespace: workflow.Namespace,
+	}
+	revisionList := &servingv1.RevisionList{}
+	if err := h.C.List(ctx, revisionList, opts); err != nil {
```
wmedvede, line 274: IDK if this reading is sorted, to be sure we get the revisions kind of
revision_1
revision_2
revision_3
...

I think this might have an impact on the iteration below.

jianrongzhang89: @wmedvede K8s LIST does not guarantee the order of the result set, so I added the code to sort the list.

wmedvede (approved):
Algorithm looks safe now, LGTM.

Many thanks, great work.

ricardozanini (approved):

